PR #46, opened 2021-03-19
Release 0.1.16

dweindl:


FFroehlich (approved):

.github/workflows/deploy.yml
```diff
@@ -14,7 +14,7 @@ jobs:
     - name: Set up python
       uses: actions/setup-python@v1
       with:
-        python-version: 3.6
+        python-version: 3.8
```
yannikschaelte, line 17: why not 9?

yannikschaelte (commented):
too late to approve, but looks good

.github/workflows/deploy.yml
dweindl: :man_shrugging:

PR #48, opened 2021-03-19
Fix override check and add noise formula check

FFroehlich:


petab/lint.py
```diff
@@ -527,53 +527,51 @@ def assert_parameter_estimate_is_boolean(parameter_df: pd.DataFrame) -> None:
...
-        if col not in measurement_df:
+    for col, allow_scalar_numeric in [
+        (OBSERVABLE_PARAMETERS, allow_scalar_numeric_observable_parameters),
+        (NOISE_PARAMETERS, allow_scalar_numeric_noise_parameters)
+    ]:
+        if col not in measurement_df or not allow_scalar_numeric:
             continue
-        measurement_df.loc[measurement_df[col].apply(is_numeric), col] = np.nan
+        measurement_df.loc[
+            measurement_df.noiseParameters.apply(isinstance,
```
dweindl, line 564:
```suggestion
            measurement_df.noiseParameters.apply(isinstance,
```

Should be `col` instead of `noiseParameters`?

petab/lint.py
```diff
@@ -598,7 +615,7 @@ def measurement_table_has_observable_parameter_numeric_overrides(
         measurement_df: PEtab measurement table
 
     Returns:
-        True if there any numbers to override observable parameters,
+        True if there any numbers to override observable/noise parameters,
```
dweindl, line 618:
```suggestion
        True if there are any numbers to override observable/noise parameters,
```

dweindl (approved):

petab/lint.py
FFroehlich: yes

PR #49, opened 2021-03-19
fix noise formula check

FFroehlich:
allowing inconsistent typing makes it much more "interesting" to implement new features.

### DIFF
--- a/petab/lint.py
+++ b/petab/lint.py
@@ -622,7 +622,8 @@ def observable_table_has_nontrivial_noise_formula(
     """
 
     return not observable_df[NOISE_FORMULA].apply(
-        lambda x: re.match(r'^[\w_\.]+$', str(x)) is not None
+        lambda x: isinstance(x, numbers.Number) or
+        re.match(r'^[\w_\.]+$', str(x)) is not None
     ).all()
 
 
--- a/petab/lint.py
+++ b/petab/lint.py
@@ -622,7 +622,7 @@ def observable_table_has_nontrivial_noise_formula(
     """
 
     return not observable_df[NOISE_FORMULA].apply(
-        lambda x: re.match(r'^[\w_\.]+$', x) is not None
+        lambda x: re.match(r'^[\w_\.]+$', str(x)) is not None
     ).all()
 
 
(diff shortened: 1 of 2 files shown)


petab/lint.py
```diff
@@ -622,7 +622,7 @@ def observable_table_has_nontrivial_noise_formula(
     """
 
     return not observable_df[NOISE_FORMULA].apply(
-        lambda x: re.match(r'^[\w_\.]+$', x) is not None
+        lambda x: re.match(r'^[\w_\.]+$', str(x)) is not None
```
dweindl, line 625: To make it more interesting, `str(x)` might be a float in scientific notation :sweat_smile: 
I think it would be easier to handle numbers separately.

tests/test_lint.py
```diff
@@ -86,11 +86,14 @@ def test_observable_table_has_nontrivial_noise_formula():
...
-        OBSERVABLE_ID: ['0obsPar1noisePar', '2obsPar0noisePar'],
+        OBSERVABLE_ID: ['0obsPar1noisePar', '2obsPar0noisePar',
+                        '3obsPar0noisePar'],
         OBSERVABLE_FORMULA: ['1.0',
+                             '1.0',
                              '1.0'],
         NOISE_FORMULA: ['noiseParameter1_0obsPar1noisePar + 3.0',
-                        '1.0']
+                        1.0,
+                        '0.5']
```
dweindl, line 96:
```suggestion
                        1e+18]
```

petab/lint.py
```diff
@@ -622,7 +623,7 @@ def observable_table_has_nontrivial_noise_formula(
     """
 
     return not observable_df[NOISE_FORMULA].apply(
-        lambda x: isinstance(x, numbers.Number) or
+        lambda x: is_scalar_float(x) or
         re.match(r'^[\w_\.]+$', str(x)) is not None
```
dweindl, line 627:
```suggestion
        re.match(r'^\w+$', str(x)) is not None
```
If numbers are handled separately, we can remove the dot. Underscore should be in \w already.

dweindl (approved):

petab/lint.py
FFroehlich: 👍

PR #51, opened 2021-03-22
Parameter mapping: Add option to ignore time-point specific noiseParameters

dweindl:
... in case those are fixed values and noiseFormula consists only of one single parameter.

### DIFF
--- a/petab/parameter_mapping.py
+++ b/petab/parameter_mapping.py
@@ -165,7 +165,7 @@ def _map_condition(packed_args):
     (condition, measurement_df, condition_df, parameter_df, sbml_model,
      simulation_parameters, warn_unmapped, scaled_parameters,
      fill_fixed_parameters,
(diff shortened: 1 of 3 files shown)


petab/parameter_mapping.py
```diff
@@ -42,6 +42,7 @@ def get_optimization_to_simulation_parameter_mapping(
         warn_unmapped: Optional[bool] = True,
         scaled_parameters: bool = False,
         fill_fixed_parameters: bool = True,
+        allow_scalar_timepoint_specific_numeric_noise_parameters: bool = False
```
yannikschaelte, line 45: this is ugly

yannikschaelte (approved):
is it nowhere used?

petab/parameter_mapping.py
dweindl: :crying_cat_face:

yannikschaelte: ALSCATISPENUMNOIPA!

PR #56, opened 2021-03-24
fix override check for empty measurement table

FFroehlich:


dweindl (approved):
Not meaningful to call that function without a measurement table, but the change doesn't hurt either. Update typehint?

yannikschaelte (approved):

yannikschaelte:
was the merge to master intended?

FFroehlich:
> was the merge to master intended?

no 🙈

paulstapor:
Still... can we have that merged soon? Need it in parPE, as it breaks CI...

PR #57, opened 2021-03-28
Optional relative paths in generated YAML

dilpath:
- feel free to suggest a better name than `yaml_centric`
- not sure if it should default to relative paths when calling `petab.Problem.to_files` and/or `petab.yaml.create_problem_yaml` (currently only the former)

### DIFF
--- a/petab/yaml.py
+++ b/petab/yaml.py
@@ -181,7 +181,7 @@ def create_problem_yaml(
         observable_files: Union[str, List[str]],
         yaml_file: str,
         visualization_files: Optional[Union[str, List[str]]] = None,
-        yaml_centric: bool = False,
+        relative_paths: bool = True,
 ) -> None:
     """
     Create and write default YAML file for a single PEtab problem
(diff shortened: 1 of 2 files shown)


petab/yaml.py
```diff
@@ -192,6 +195,8 @@ def create_problem_yaml(sbml_files: Union[str, List[str]],
         yaml_file: Path to which YAML file should be written
         visualization_files: Optional Path to visualization file or list of
         such
+        yaml_centric: whether all paths in the YAML file should be relative to
```
yannikschaelte, line 198: also mention the alternative

petab/yaml.py
```diff
@@ -204,6 +209,20 @@ def create_problem_yaml(sbml_files: Union[str, List[str]],
     if isinstance(visualization_files, str):
         visualization_files = [visualization_files]
 
+    if yaml_centric:
+        yaml_file_dir = Path(yaml_file).parent
+
+        def get_relative_paths(paths, start=yaml_file_dir):
+            return [os.path.relpath(path, start=start) for path in paths]
```
yannikschaelte, line 216:
```suggestion
        def get_rel_to_yaml(paths: List[str]):
            return [os.path.relpath(path, start=yaml_file_dir) for path in paths]
```

yannikschaelte (approved):

dweindl:
> * feel free to suggest a better name than `yaml_centric`

`relpath` (in line with the respective Python function), `relative_paths`, `relative`?

> * not sure if it should default to relative paths when calling `petab.Problem.to_files` and/or `petab.yaml.create_problem_yaml` (currently only the former)

I think it makes sense to change the default to True. It's a breaking change, but I wouldn't expect many users to rely on having absolute paths in there.

dweindl (approved):

PR #60, opened 2021-04-15
Fix unreachable TypeError in petab.parameter_mapping.apply_overrides_…

### LINKED ISSUE #59: Unreachable TypeError in `petab.parameter_mapping.apply_overrides_for_observable`
I don't think https://github.com/PEtab-dev/libpetab-python/blob/0917194199df10b40e3c16dd567a1fdcda0c08c7/petab/parameter_mapping.py#L365 can ever be reached.

If `overridee_id` is not present in `mapping`, the code will simply add a new entry and not throw a `KeyError`.

dweindl:
…for_observable (fixes #59)

Revert "More informative error messages in case of wrongly set observable and noise parameters (Closes PEtab-dev/PEtab#118) (PEtab-dev/PEtab#155)"

This reverts commit 8fab85e2b581ff501ee0a5595faa4d28543c0257.

### DIFF
--- a/petab/parameter_mapping.py
+++ b/petab/parameter_mapping.py
@@ -360,14 +360,7 @@ def _apply_overrides_for_observable(
     """
     for i, override in enumerate(overrides):
         overridee_id = f'{override_type}Parameter{i+1}_{observable_id}'
-        try:
-            mapping[overridee_id] = override
-        except KeyError as e:
-            raise TypeError(f'Cannot override {override_type} parameter '
-                            f'{overridee_id} for observable {observable_id}.'
-                            f'Ensure there exists an {override_type} '
-                            'definition containing the correct number of '
-                            'placeholder parameters.') from e
+        mapping[overridee_id] = override
 
 
 def _apply_condition_parameters(par_mapping: ParMappingDict,


FFroehlich:
I think it would actually be nice to have an error message when mapping is not possible.

dweindl:
> I think it would actually be nice to have an error message when mapping is not possible.

Such a case would be invalid PEtab and should be reported by `petablint`. In the parameter mapping we widely assume we have valid PEtab files. If you have an example where this is not correctly detected, please share.

FFroehlich:
> > I think it would actually be nice to have an error message when mapping is not possible.
> 
> Such a case would be invalid PEtab and should be reported by `petablint`. In the parameter mapping we widely assume we have valid PEtab files. If you have an example where this is not correctly detected, please share.

This is not detected for pysb-petab. Used `observableParameters_$obs_id$` (extra s) or misspelled obs_id and both definitely did not yield the results I wanted.

yannikschaelte (approved):

dweindl:
> This is not detected for pysb-petab. Used `observableParameters_$obs_id$` (extra s) or misspelled obs_id and both definitely did not yield the results I wanted.

Hm.... changing petab test case 0003:
```
diff --git a/petabtests/cases/pysb/0003/_observables.tsv b/petabtests/cases/pysb/0003/_observables.tsv
index dc2a127..43e4a45 100644
--- a/petabtests/cases/pysb/0003/_observables.tsv
+++ b/petabtests/cases/pysb/0003/_observables.tsv
@@ -1,2 +1,2 @@
 observableId   observableFormula       noiseFormula
-obs_a  observableParameter1_obs_a * A + observableParameter2_obs_a     0.5
+obs_a  observableParameter1_obs_a * A + observableParameters2_obs_a    0.5
``` 
gives me

```
petablint -y 0003/_0003.yaml 
SBML model not available. Skipping.
Mismatch of observable parameter overrides for obs_a (observableParameter1_obs_a * A + observableParameters2_obs_a)in:
observableId             obs_a
simulationConditionId       c0
time                         0
measurement                0.7
observableParameters     0.5;2
Name: 0, dtype: object
Expected 0 or 1 but got 2
Not OK
```

What am I doing differently?

PR #68, opened 2021-07-05
Fix erroneous bitwise comparison in petab.visualize.DataProvider

dweindl:


petab/visualize/plotting.py
```diff
@@ -476,7 +476,7 @@ def get_data_series(self, data_df: pd.DataFrame, data_col: str,
             measurements_to_plot.at[var_cond_id, 'sd'] = np.std(
                 data_measurements)
 
-            if provided_noise & sum(subset):
+            if provided_noise and sum(subset):
```
dweindl, line 479:
```suggestion
            if provided_noise and np.any(subset):
```

@plakrisenko : This is what is meant here, right?

plakrisenko: This was simply transferred from https://github.com/PEtab-dev/libpetab-python/blob/master/petab/visualize/helper_functions.py#L981

plakrisenko: But I think you're right

plakrisenko (approved):

Issue #61, opened 2021-04-23
lint doesnt check for `;` at the end of observable parameter definitions


FFroehlich:
simply add ';' at the end of an `petab.OBSERVABLE_PARAMETERS`, should ideally lead to a check for `''` in the parameter table which fails

PR #69, opened 2021-07-08
Lint: Check for valid parameter IDs in observable/noise parameters

dweindl:
Closes PEtab-dev/libpetab-python#61

### DIFF
--- a/petab/measurements.py
+++ b/petab/measurements.py
@@ -221,10 +221,10 @@ def split_parameter_replacement_list(
             return x
         if lint.is_valid_identifier(x):
             return x
+
         raise ValueError(
-            f"The value '{x}' (whitespace removed) in the parameter "
-            f"replacement list '{list_string}' is neither a number, nor a "
-            "valid parameter ID."
+            f"The value '{x}' in the parameter replacement list "
+            f"'{list_string}' is neither a number, nor a valid parameter ID."
         )
 
     return list(map(convert_and_check, result))
--- a/petab/measurements.py
+++ b/petab/measurements.py
@@ -221,9 +221,11 @@ def split_parameter_replacement_list(
             return x
         if lint.is_valid_identifier(x):
             return x
-        raise ValueError(f"'{x}' in parameter replacement list "
-                         f"'{list_string}' is neither a number, nor a "
-                         "valid parameter ID.")
+        raise ValueError(
+            f"The value '{x}' (whitespace removed) in the parameter "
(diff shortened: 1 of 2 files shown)


petab/measurements.py
```diff
@@ -200,9 +200,10 @@ def split_parameter_replacement_list(
         delim: delimiter
 
     Returns:
-         List of split values. Numeric values converted to float.
+         List of split values. Numeric values converted to numeric type,
+         parameter IDs as strings.
     """
-    if list_string is None:
+    if list_string is None or list_string == '':
```
dilpath, line 206:
```suggestion
    if list_string is None or list_string.strip() == '':
```

dilpath: Ah, would need to move after the `isinstance` check below though.

dweindl: That was actually intentional. I would consider blanks there an error.

petab/measurements.py
```diff
@@ -200,9 +200,10 @@ def split_parameter_replacement_list(
         delim: delimiter
 
     Returns:
-         List of split values. Numeric values converted to float.
+         List of split values. Numeric values converted to numeric type,
+         parameter IDs as strings.
```
dilpath, line 204:
```suggestion
         List of split values. Numeric values may be converted to `float`,
         and parameter IDs are kept as strings.
```

petab/measurements.py
```diff
@@ -212,8 +213,19 @@ def split_parameter_replacement_list(
...
+
+    def convert_and_check(x):
+        x = core.to_float_if_float(x)
+        if isinstance(x, float):
+            return x
+        if lint.is_valid_identifier(x):
+            return x
+        raise ValueError(f"'{x}' in parameter replacement list "
+                         f"'{list_string}' is neither a number, nor a "
+                         "valid parameter ID.")
```
dilpath, line 226: `x` has had `strip()` applied to it, so may be confusing to see possibly just `''`.

```suggestion
        raise ValueError(
            f"The value '{x}' (whitespace removed) in the parameter "
            f"replacement list '{list_string}' is neither a number, nor a "
            "valid parameter ID."
        )
```

dilpath (approved):
:+1:

petab/measurements.py
dweindl: Ah, will remove the `strip()`. We don't want that here.

plakrisenko (approved):

PR #76, opened 2021-10-01
Ensure measurements are numeric and not null with linter

dilpath:
A benchmark model was passing the linter despite having null measurement values (see https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab/pull/129).

### DIFF
--- a/petab/lint.py
+++ b/petab/lint.py
@@ -20,6 +20,8 @@ logger = logging.getLogger(__name__)
 __all__ = ['assert_all_parameters_present_in_parameter_df',
            'assert_measured_observables_defined',
            'assert_measurement_conditions_present_in_condition_table',
+           'assert_measurements_not_null',
+           'assert_measurements_numeric',
            'assert_model_parameters_in_condition_or_parameter_table',
            'assert_no_leading_trailing_whitespace',
            'assert_noise_distributions_valid',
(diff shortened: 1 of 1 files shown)


petab/lint.py
```diff
@@ -177,6 +179,9 @@ def check_measurement_df(df: pd.DataFrame,
         measurements.assert_overrides_match_parameter_count(
             df, observable_df)
 
+    assert_measurements_not_null(df)
```
dweindl, line 182: At some point, empty/NaN measurements were intentionally allowed. I think it was used to mark additional output timepoints for visualization. Not sure about any other use cases. 
Although not having NaNs there is probably a good idea, I'd be happy to hear additional opinions (e.g. @yannikschaelte )

dilpath: Agreed; however, the spec currently requires `NOT NULL`: https://petab.readthedocs.io/en/latest/documentation_data_format.html#id1

A drawback would be that counting the number of measurements would be slightly less intuitive, e.g.
- https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab/blob/master/Benchmark-Models/model_info.py#L24-L25
- computing the Bayesian information criterion

An alternative, if measurements are `NOT NULL`, would be for users to maintain "simulation-only" measurements in a second measurements file, with dummy values, and only use it for simulation (not optimization) (so, two problem YAMLs).

PR #83, opened 2021-11-07
allow zero bounds for log parameters

FFroehlich:
if there is no finiteness check for upper bound, I don't see why we should prevent lower bound 0 for log scale parameter, it convenient for specifying unbounded parameters after all.


yannikschaelte (approved):

dweindl (approved):
Agreed

PR #90, opened 2021-11-25
Visualization: Don't mess with rcParams

dweindl:
Setting matplotlib rcParams makes it impossible to use custom styles, and therefore, should best be avoided.

### DIFF
--- a/petab/visualize/plotter.py
+++ b/petab/visualize/plotter.py
@@ -340,23 +340,13 @@ class MPLPlotter(Plotter):
             In case subplots are saved to file.
         """
 
-        # Set Options for plots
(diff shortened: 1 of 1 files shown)


plakrisenko:
I guess the appearance is ok with the defaults, but should we also provide some custom styles, e.g. using  'tableau-colorblind10' and one suitable for presentation?

plakrisenko (approved):

dweindl:
> should we also provide some custom styles, e.g. using 'tableau-colorblind10' and one suitable for presentation?

I would keep the matplotlib defaults and let the user choose their style. The proper style depends on many factors... Fine to use that color palette in the example notebooks.

PR #92, opened 2021-11-29
Linter: Check condition IDs are unique

### LINKED ISSUE #70: no lint error for non-unique condition ids
`petab.lint.check_condition_df` does not check whether index entries are unique, but this is required according to documentation.

dweindl:
Closes #70

### DIFF
--- a/petab/lint.py
+++ b/petab/lint.py
@@ -108,6 +108,10 @@ def check_condition_df(
 
     check_ids(df.index.values, kind='condition')
 
+    if not df.index.is_unique:
+        raise AssertionError("Non-unique condition IDs: "
+                             f"{df.index.values[df.index.duplicated()]}")
+
     for column_name in req_cols:
         if not np.issubdtype(df[column_name].dtype, np.number):
             assert_no_leading_trailing_whitespace(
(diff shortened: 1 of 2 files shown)


petab/lint.py
```diff
@@ -108,6 +108,10 @@ def check_condition_df(
 
     check_ids(df.index.values, kind='condition')
 
+    if not df.index.is_unique:
```
dilpath, line 111: Maybe this is inappropriate according to the docs, and could be replaced with `if df.index.has_duplicates`
```python
>>> print(pd.Index.is_unique.__doc__)

        Return if the index has unique values.
        
>>> print(pd.Index.has_duplicates.__doc__)

        Check if the Index has duplicate values.
```

dilpath: The tests may have passed because there are always only duplicated or only unique values in any tested index (no mixture).

dilpath (approved):
:+1:

petab/lint.py
dweindl: I think the docstring there maybe slightly misleading. `has_duplicates` is only `return not self.is_unique `. So, I'd directly use the latter.

PR #96, opened 2021-11-30
Fix petab.visualize.data_overview.create_report

dweindl:
* Fix number of conditions
* Handle optional preequilibrationConditionId

### DIFF
--- a/petab/visualize/data_overview.py
+++ b/petab/visualize/data_overview.py
@@ -31,8 +31,7 @@ def create_report(
     template_file = "report.html"
 
     data_per_observable = get_data_per_observable(problem.measurement_df)
-    num_conditions = (len(problem.condition_df.columns)
-                      - 1 * (CONDITION_NAME in problem.condition_df.columns))
+    num_conditions = len(problem.condition_df.index)
 
     # Setup template engine
     import jinja2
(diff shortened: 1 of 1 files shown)


petab/visualize/data_overview.py
```diff
@@ -55,13 +54,14 @@ def get_data_per_observable(measurement_df: pd.DataFrame) -> pd.DataFrame:
...
+        Pivot table with number of data points per observable and condition
     """
 
     my_measurements = measurement_df.copy()
-
-    my_measurements[PREEQUILIBRATION_CONDITION_ID].fillna('', inplace=True)
+    if PREEQUILIBRATION_CONDITION_ID not in my_measurements:
+        my_measurements[PREEQUILIBRATION_CONDITION_ID] = ''
+    else:
+        my_measurements[PREEQUILIBRATION_CONDITION_ID].fillna('', inplace=True)
```
dilpath, line 64:
```suggestion
    index = [SIMULATION_CONDITION_ID]
    if PREEQUILIBRATION_CONDITION_ID in my_measurements:
        my_measurements[PREEQUILIBRATION_CONDITION_ID].fillna('', inplace=True)
        index.append(PREEQUILIBRATION_CONDITION_ID)
```
then `index=index` below, to avoid unspecified optional columns in the result?

dweindl: thx, yes

PR #108, opened 2022-02-11
Generation of condition-specific SBML models

dweindl:
New function `petab.get_model_for_condition` to generate condition-specific SBML models based on a PEtab problem.

### DIFF
--- a/petab/sbml.py
+++ b/petab/sbml.py
@@ -548,8 +548,7 @@ def get_model_for_condition(
     def remove_rules(target_id: str):
         if sbml_model.removeRuleByVariable(target_id):
             warn("An SBML rule was removed to set the component "
-                 f"{sbml_model.getElementBySId(target_id).getId()} to a "
-                 "constant value.")
+                 f"{target_id} to a constant value.")
         sbml_model.removeInitialAssignment(target_id)
 
     for parameter in sbml_model.getListOfParameters():
--- a/petab/sbml.py
+++ b/petab/sbml.py
@@ -571,5 +571,28 @@ def get_model_for_condition(
             sbml_species.setInitialConcentration(new_value)
         elif sbml_species.isSetInitialAmount():
             sbml_species.setInitialAmount(new_value)
+        else:
+            raise AssertionError(
+                "Neither initial amount not initial concentration "
+                f"set for {sbml_species.getId()}")
+
+    # set compartment size for any compartments in the condition table
+    for component_id in petab_problem.condition_df:
+        sbml_compartment = sbml_model.getCompartment(component_id)
+        if not sbml_compartment:
+            continue
+
+        # if there is an initial assignment for that compartment, remove it,
+        #  as the compartment in the condition table would override it
+        sbml_model.removeInitialAssignment(component_id)
+
+        # set initial concentration/amount
+        new_value = petab.to_float_if_float(
+            petab_problem.condition_df.loc[sim_condition_id, component_id])
+        if not isinstance(new_value, Number):
+            # parameter reference in condition table
+            new_value = get_param_value(new_value)
+
+        sbml_compartment.setSize(new_value)
 
     return sbml_doc, sbml_model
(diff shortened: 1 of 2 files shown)


dilpath (approved):
:+1:

Looks like compartments in a conditions table aren't yet handled

dweindl:
> Looks like compartments in a conditions table aren't yet handled

Oops. Thanks :).

petab/sbml.py
```diff
@@ -481,3 +485,91 @@ def load_sbml_from_file(
...
+        new_value = petab.to_float_if_float(
+            petab_problem.condition_df.loc[sim_condition_id, condition_par_id])
+        if not isinstance(new_value, Number):
+            # parameter reference in condition table
+            new_value = get_param_value(new_value)
+
+        if sbml_species.isSetInitialConcentration():
+            sbml_species.setInitialConcentration(new_value)
+        elif sbml_species.isSetInitialAmount():
+            sbml_species.setInitialAmount(new_value)
```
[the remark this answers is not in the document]
dweindl: One of those two has to be set to be valid SBML iirc. But let's raise to be on the safe side.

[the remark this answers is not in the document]
dilpath: Apparently fine if e.g. initial assignment is used. From SBML spec:
> Missing initialAmount and initialConcentration
values implies that their values either are unknown, or to be obtained from an external source, or determined
by an initial assignment (Section 4.8 on p. 56) or other SBML construct elsewhere in the model.

petab/sbml.py
```diff
@@ -481,3 +485,114 @@ def load_sbml_from_file(
...
+        else:
+            raise AssertionError(
+                "Neither initial amount not initial concentration "
```
dilpath, line 576:
```suggestion
                "Neither initial amount nor initial concentration "
```

dilpath (approved):
:+1:

Another source of initial values is from assignment rules and algebraic rules. From SBML spec:
> If the species’ constant attribute is “ false ”, the species’ quantity value may be overridden by an InitialAssignment or changed by AssignmentRule or AlgebraicRule [...]

Doesn't make sense to override such rules just to set an initial value, but maybe such a contradiction (e.g. compartment in the condition table, but also with an assignment rule) could be identified by `petablint`.

dweindl:
> Another source of initial values is from assignment rules and algebraic rules. [...]

True. It's currently not well defined what should happen in this case (see also discussion in https://github.com/PEtab-dev/PEtab/issues/513). Let's remove all rules targeting replaced entities for now.

petab/sbml.py
```diff
@@ -481,3 +485,119 @@ def load_sbml_from_file(
...
+        # estimated parameter, look up in nominal parameters
+        return petab_problem.parameter_df.loc[mapped_value,
+                                              petab.NOMINAL_VALUE]
+
+    for parameter in sbml_model.getListOfParameters():
+        new_value = get_param_value(parameter.getId())
+        if new_value:
+            parameter.setValue(new_value)
+            # remove rules that would override that value
+            sbml_model.removeRuleByVariable(parameter.getId())
```
dilpath, line 553:
```suggestion
            if sbml_model.removeRuleByVariable(parameter.getId()) is not None:
                warnings.warn(f'An SBML rule was removed to set the component {parameter.getId()} to a constant value.')
```
Since I guess this may more often be an error than intended behavior.

petab/sbml.py
```diff
@@ -481,3 +485,119 @@ def load_sbml_from_file(
...
+
+        # estimated parameter, look up in nominal parameters
+        return petab_problem.parameter_df.loc[mapped_value,
+                                              petab.NOMINAL_VALUE]
+
+    def remove_rules(target_id: str):
+        if sbml_model.removeRuleByVariable(target_id):
+            warn("An SBML rule was removed to set the component "
+                 f"{sbml_model.getElementBySId(target_id).getId()} to a "
+                 "constant value.")
```
dilpath, line 552:
```suggestion
                 f"{target_id} to a constant value.")
```

dweindl: d'oh

PR #110, opened 2022-03-01
Less ambiguous naming in generic output

dilpath:
Not really intended use case, but e.g. a PEtab Select problem may exist in the same directory.


dweindl:
Hm. This might easily break things for people already using that function. Not sure I like changing the yaml file name without pressing need.

dilpath:
> Hm. This might easily break things for people already using that function. Not sure I like changing the yaml file name without pressing need.

Yes, will probably break things in tools that interact with PEtab Select via its CLI...

PR #114, opened 2022-03-25
Add steady state constant

dilpath:


petab/C.py
```diff
@@ -20,6 +20,9 @@
 #:
 TIME = 'time'
 
+#: Time value that indicates steady-state measurements
+STEADY_STATE = 'inf'
```
dweindl, line 24: What about using `TIME_STEADY_STATE`? Would be clearer that it doesn't refer to the steady-state as such?

dweindl: Not completely sure whether one would expect a string or a number here. I see use cases for either.

dweindl (approved):

dweindl (approved):

PR #115, opened 2022-03-25
fix jinja version

dilpath:
- https://github.com/jupyter/nbconvert/issues/1736
- https://github.com/AMICI-dev/AMICI/pull/1742


dweindl (approved):
Thanks

PR #58, opened 2021-04-06
Visualization refactoring

plakrisenko:
Major visualization refactoring. 
Visualization functionality is rewritten using object oriented approach. 

It was decided to get rid of grouping by number options (`sim_cond_num_list`, `observable_num_list` parameters in `plot_data_and_simulation`) as it doesn't seem to be clear that the numbering is based on the order in the measurement_df. 

Additionally, it is now possible to set axes limits for all subplots (`Figure.set_axes_limits`). 

Fixed issues:
-  noise visualization issues, plotting replicates: #31, #33, #24
-  barplot labeling issues: #23 
-  plotting only simulation: #13 
-  plotting bug for the case when measurements order is different in measurements_df and simulations_df: #30 and #36
-  vis spec expansion issue: #28, #34 
-  yValues order: #38

PR #130, opened 2022-04-27
Remove deprecated visualization functions

dweindl:
Remove old visualization functions deprecated since #58.


dweindl:
The example notebooks still use the old vis code?! :(

plakrisenko (approved):

PR #133, opened 2022-04-28
Add abstraction for (SBML) models

dweindl:
To accommodate https://github.com/PEtab-dev/PEtab/pull/538

### DIFF
--- a/petab/models/sbml_model.py
+++ b/petab/models/sbml_model.py
@@ -1,7 +1,7 @@
 """Functions for handling SBML models"""
 
 from .model import Model
-from ..sbml import get_sbml_model
+from ..sbml import get_sbml_model, load_sbml_from_string
 import libsbml
 from typing import Optional, Iterable, Tuple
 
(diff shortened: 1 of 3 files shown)


petab/lint.py
```diff
@@ -764,13 +762,11 @@ def lint_problem(problem: 'petab.Problem') -> bool:
...
 
     # Run checks on individual files
-    if problem.sbml_model is not None:
-        logger.info("Checking SBML model...")
-        errors_occurred |= not sbml.is_sbml_consistent(
-            problem.sbml_model.getSBMLDocument())
-        sbml.log_sbml_errors(problem.sbml_model.getSBMLDocument())
+    if problem.model is not None:
+        logger.info("Checking model...")
+        errors_occurred |= not problem.model.is_valid()
```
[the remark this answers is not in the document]
dweindl: Yes, it's the same. Would leave it as it is to be more robust about careless rearrangements :grimacing:

petab/models/__init__.py
```diff
@@ -0,0 +1,5 @@
+MODEL_TYPE_SBML = 'sbml'
+
+known_model_types = {MODEL_TYPE_SBML}
```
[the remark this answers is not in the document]
dweindl: I'd prefer keeping it more modular, but could consider importing it in `petab.C`

petab/models/sbml_model.py
```diff
@@ -0,0 +1,120 @@
...
+        sbml_reader, sbml_document, sbml_model = get_sbml_model(
+            filepath_or_buffer)
+        return SbmlModel(
+            sbml_model=sbml_model,
+            sbml_reader=sbml_reader,
+            sbml_document=sbml_document,
+        )
+
+    def to_file(self, filename: [str, Path]):
+        from ..sbml import write_sbml
```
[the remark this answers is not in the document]
dweindl: Good question

petab/measurements.py
```diff
@@ -154,7 +154,7 @@ def get_unique_parameters(series):
 
 def split_parameter_replacement_list(
         list_string: Union[str, numbers.Number],
-        delim: str = ';') -> List[Union[str, numbers.Number]]:
+        delim: str = PARAMETER_SEPARATOR) -> List[Union[str, numbers.Number]]:
```
FFroehlich, line 157: does it ever make sense to pass anything but `PARAMETER_SEPARATOR` here?

dweindl: Probably not for now. *If* the separator ever changes, it can be useful, but shouldn't hurt keeping the optional argument.

FFroehlich: but why is it optional for this specific function (in contrast to all the other code, where it isn't optional)?

dweindl: Because adding it to the whole potential call stack will be a mess :see_no_evil:

FFroehlich: 🥇

Issue #89, opened 2021-11-25
Update visualization notebooks


dweindl:
Probably didn't survive the recent visualization refactoring...

plakrisenko:
What exactly doesn't work? 
The old code wasn't removed yet, so they should be fine

dweindl:
```
---------------------------------------------------------------------------

NotImplementedError                       Traceback (most recent call last)

/tmp/ipykernel_509298/4631237.py in <module>
----> 1 ax_without_sim = plot_data_and_simulation(
      2     data_file_path,
      3     condition_file_path,
      4     visualization_file_path)
      5 plt.show()

~/src/libpetab-python/petab/visualize/plot_data_and_simulation.py in plot_data_and_simulation(exp_data, exp_conditions, vis_spec, sim_data, dataset_id_list, sim_cond_id_list, sim_cond_num_list, observable_id_list, observable_num_list, plotted_noise, subplot_file_path)
    179         for _, plot_spec in vis_spec[ind_plot].iterrows():
    180             # handle plot of current dataset
--> 181             handle_dataset_plot(plot_spec, ax, exp_data,
    182                                 exp_conditions, sim_data)
    183 

~/src/libpetab-python/petab/visualize/helper_functions.py in handle_dataset_plot(plot_spec, ax, exp_data, exp_conditions, sim_data)
    890     nan_set = all([np.isnan(val) for val in measurement_to_plot['mean']])
    891     if not nan_set:
--> 892         plot_lowlevel(plot_spec, ax, conditions, measurement_to_plot, plot_sim)
    893 
    894     # Beautify plots

~/src/libpetab-python/petab/visualize/plotting_config.py in plot_lowlevel(plot_spec, ax, conditions, ms, plot_sim)
    158     elif plot_spec[PLOT_TYPE_SIMULATION] == SCATTER_PLOT:
    159         if not plot_sim:
--> 160             raise NotImplementedError('Scatter plots do not work without'
    161                                       ' simulation data')
    162         ax.scatter(ms['mean'], ms['sim'],

NotImplementedError: Scatter plots do not work without simulation data

```

```
---------------------------------------------------------------------------

IndexError                                Traceback (most recent call last)

/tmp/ipykernel_635809/2240385560.py in <module>
      6                               visualization_files=visualization_file_path)
      7 
----> 8 ax = petab.visualize.plot_petab_problem(petab_problem=pp)
      9 

[...]

PR #137, opened 2022-04-28
Update first vis example notebook; use non-deprecated functions

dweindl:
See #89

Also, clear notebook output to not litter the repository - will be run during doc build (#144).


plakrisenko (approved):

plakrisenko:
Only small thing: the plots with an empty VS and with only 'plotId' column aren't the same as stated. In the former case there are three subplots and in the latter case everything is plotted together.

dweindl:
> Only small thing: the plots with an empty VS and with only 'plotId' column aren't the same as stated. In the former case there are three subplots and in the latter case everything is plotted together.

Thanks. Updated.

PR #139, opened 2022-04-28
Simplify test model creation - use simplesbml

dweindl:

PR #140, opened 2022-04-28
Remove ancient deprecated functions related to specifying observables/noise models inside SBML

dweindl:
They are obsolete since the introduction of the observable table.


dweindl:
- [ ] #139

FFroehlich (approved):

PR #141, opened 2022-04-29
Refactor yaml validation schemas

dweindl:
In the future, we want to include separate validation schemas for each PEtab version in this package


FFroehlich (approved):

petab/yaml.py
```diff
@@ -52,13 +58,17 @@ def validate_yaml_syntax(
...
+    yaml_config = load_yaml(yaml_config)
 
     if schema is None:
-        schema = SCHEMA
+        # try get PEtab version from yaml file
+        #  if this is not the available, the file is not valid anyways,
+        #  but let's still use the latest PEtab schema for full validation
+        version = yaml_config.get(FORMAT_VERSION, None) \
+                      or list(SCHEMAS.values())[-1]
+        schema = SCHEMAS[str(version)]
```
FFroehlich, line 69: shouldn't this include some fallback if version is None?

PR #150, opened 2022-05-19
Fix handling of NaN values for parameters in condition table

dweindl:
In such cases, the parameter value from the model should be used.
So far, NaN was kept.

### DIFF
--- a/petab/parameter_mapping.py
+++ b/petab/parameter_mapping.py
@@ -410,7 +410,8 @@ def _apply_condition_parameters(par_mapping: ParMappingDict,
         par_mapping[overridee_id] = core.to_float_if_float(
             condition_df.loc[condition_id, overridee_id])
 
-        if np.isnan(par_mapping[overridee_id]):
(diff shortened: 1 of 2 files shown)


FFroehlich (approved):
👍

FFroehlich:
Is there a petab testsuite case that covers condition tables with `NaNs`?

dweindl:
> Is there a petab testsuite case that covers condition tables with `NaNs`?

Yes, but only for species, not for parameters.

leodarrigade:
Hi all,
it is my first time trying to use PEtab and PyPESTO. I have a question regarding this issue. Say I have a parameter k that I want to estimate for certain experimental conditions (condition A) and to be fixed (k=1) for other experimental conditions (condition B). As far as I understand, a way to pass this information to PEtab would be :
- in parameters table, set estimate = 1 (True) for k
- in conditions table, in column k, for condition A write Nan and for condition B write 1.
Therefore, in condition A, the algorithm will switch to what is described in parameters table and will estimate k.

Am I correct ?

Best

PR #152, opened 2022-05-24
Add support for PEtab problems with multiple condition files

### LINKED ISSUE #7: Add support for multiple condition files per petab.Problem
Allowed in YAML file, but not implemented in library.

dweindl:
Allows specifying different conditions in different files.

Closes #7

### DIFF
--- a/petab/problem.py
+++ b/petab/problem.py
@@ -88,7 +88,8 @@ class Problem:
     @staticmethod
     def from_files(
             sbml_file: Union[str, Path, None] = None,
-            condition_file: Union[str, Path, None] = None,
+            condition_file:
+            Union[str, Path, None, Iterable[Union[str, Path]]] = None,
             measurement_file: Union[str, Path,
                                     Iterable[Union[str, Path]]] = None,
             parameter_file: Union[str, Path,
@@ -115,7 +116,8 @@ class Problem:
         observable_df = None
 
         if condition_file:
-            condition_df = conditions.get_condition_df(condition_file)
+            condition_df = core.concat_tables(condition_file,
+                                              conditions.get_condition_df)
 
         if measurement_file:
             # If there are multiple tables, we will merge them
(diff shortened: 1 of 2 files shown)


tests/test_petab.py
```diff
@@ -452,6 +455,34 @@ def test_concat_measurements():
...
+        'par2': [2.3],
+    }).set_index(CONDITION_ID)
+
+    df_expected = pd.DataFrame(data={
+        CONDITION_ID: ['condition1', 'condition2', 'condition3'],
+        'par1': [1.1, 1.2, 1.3],
+        'par2': [2.1, 2.2, 2.3],
+        'par3': [3.1, 3.2, np.nan],
+    }).set_index(CONDITION_ID)
+    print(petab.concat_tables((df1, df2), petab.get_condition_df))
```
dilpath, line 480:
```suggestion
```

petab/problem.py
```diff
@@ -88,7 +88,8 @@ def __setstate__(self, state):
     @staticmethod
     def from_files(
             sbml_file: Union[str, Path, None] = None,
```
dilpath, line 90:
```suggestion
            sbml_file: Union[str, Path] = None,
```
Not sure, why can `sbml_file` and `condition_file` be specified as `None` by the user, but not the other tables?

petab/problem.py
```diff
@@ -88,7 +88,8 @@ def __setstate__(self, state):
...
-            condition_file: Union[str, Path, None] = None,
+            condition_file:
+            Union[str, Path, None, Iterable[Union[str, Path]]] = None,
```
dilpath, line 92: Unrelated, but `parameter_file` can also be `Iterable[Union[str, Path]]` here, but is `List[str]` in `get_parameter_df` [1]. I guess it can be changed to `Iterable[Union[str, Path]]` at [1] too.

[1] https://github.com/PEtab-dev/libpetab-python/blob/c457c05172234ff0998eb3054b495f472fd9d0c3/petab/parameters.py#L31=

petab/problem.py
```diff
@@ -198,7 +200,10 @@ def from_yaml(yaml_config: Union[Dict, Path, str]) -> 'Problem':
 
         problem0 = yaml_config['problems'][0]
 
-        yaml.assert_single_condition_and_sbml_file(problem0)
+        if len(problem0[SBML_FILES]) > 1:
+            # TODO https://github.com/ICB-DCM/PEtab/issues/188
```
dilpath, line 204:
```suggestion
            # TODO https://github.com/PEtab-dev/libpetab-python/issues/6
```

dilpath (approved):
:+1:

petab/problem.py
dweindl: All of them should be optional.

petab/problem.py
dweindl: Will address separately

PR #159, opened 2022-07-01
Argument forwarding for Problem.get_optimization_to_simulation_parameter_mapping

dweindl:
Previously, not all options relevant options of `petab.get_optimization_to_simulation_parameter_mapping` were accessible via `Problem.get_optimization_to_simulation_parameter_mapping`. Now all keyword arguments are forwarded.


petab/problem.py
```diff
@@ -700,14 +700,11 @@ def get_simulation_conditions_from_measurement_df(self):
...
-            self,
-            warn_unmapped: bool = True,
-            scaled_parameters: bool = False,
-            allow_timepoint_specific_numeric_noise_parameters:
-            bool = False,
+            self, **kwargs
     ):
         """
-        See get_simulation_to_optimization_parameter_mapping.
+        See :py:func:`get_simulation_to_optimization_parameter_mapping`, to
```
dilpath, line 706: Where is this? Link is also broken

https://petab--159.org.readthedocs.build/projects/libpetab-python/en/159/build/_autosummary/petab.problem.html#petab.problem.Problem.get_optimization_to_simulation_parameter_mapping

dilpath (approved):

petab/problem.py
dweindl: Oops. Looks like some names changed there...

PR #165, opened 2022-07-20
Fix typehints and deprecation warning

### LINKED ISSUE #164: incorrect deprecation warning
https://github.com/PEtab-dev/libpetab-python/blob/286494a755ace31cad297076b9664023d7118d1c/petab/problem.py#L72

says I should import `petab.models.SbmlModel` but

https://github.com/PEtab-dev/libpetab-python/blob/286494a755ace31cad297076b9664023d7118d1c/petab/models/__init__.py#L5

doesn't import SbmlModel, so the class is only available from `petab.models.sbml_model.SbmlModel`

dweindl:
Closes #164

### DIFF
--- a/petab/core.py
+++ b/petab/core.py
@@ -20,7 +20,7 @@ __all__ = ['get_simulation_df', 'write_simulation_df', 'get_visualization_df',
            'create_combine_archive', 'unique_preserve_order']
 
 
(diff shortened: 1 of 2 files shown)


FFroehlich (approved):
👍

PR #172, opened 2022-08-03
Simplification of PEtab problems

dweindl:
Adds some functions to simplify / clean existing PEtab problems:
* removing unused conditions
* moving parameters that aren't really condition-dependent from the conditions table to the parameters table

### DIFF
--- a/petab/simplify.py
+++ b/petab/simplify.py
@@ -64,7 +64,7 @@ def condition_parameters_to_parameter_table(problem: Problem):
 
     replacements = {}
     for parameter_id in problem.condition_df:
-        if parameter_id == PARAMETER_NAME:
+        if parameter_id == CONDITION_NAME:
             continue
 
         if problem.model.is_state_variable(parameter_id):
--- a/petab/simplify.py
+++ b/petab/simplify.py
@@ -1,0 +1,106 @@
+"""Functionality for simplifying PEtab problems"""
+from math import nan
+
+import pandas as pd
+
+from . import Problem
+from .C import *  # noqa: F403
+from .lint import lint_problem
+
+__all__ = [
+    "remove_nan_measurements",
+    "remove_unused_observables",
+    "remove_unused_conditions",
+    "simplify_problem",
+    "condition_parameters_to_parameter_table",
+]
+
+
+def remove_nan_measurements(problem: Problem):
+    """Drop any measurements that are NaN"""
+    problem.measurement_df = problem.measurement_df[
+        ~problem.measurement_df[MEASUREMENT].isna()
+    ]
+    problem.measurement_df.reset_index(inplace=True, drop=True)
+
+
+def remove_unused_observables(problem: Problem):
+    """Remove observables that have no measurements"""
+    measured_observables = set(problem.measurement_df[OBSERVABLE_ID].unique())
+    problem.observable_df = problem.observable_df[
+        problem.observable_df.index.isin(measured_observables)]
+
+
+def remove_unused_conditions(problem: Problem):
+    """Remove conditions that have no measurements"""
+    measured_conditions = \
+        set(problem.measurement_df[SIMULATION_CONDITION_ID].unique())
+    if PREEQUILIBRATION_CONDITION_ID in problem.measurement_df:
+        measured_conditions |= \
+            set(problem.measurement_df[PREEQUILIBRATION_CONDITION_ID].unique())
+
+    problem.condition_df = problem.condition_df[
+        problem.condition_df.index.isin(measured_conditions)]
+
+
+def simplify_problem(problem: Problem):
+    if lint_problem(problem):
+        raise ValueError("Invalid PEtab problem supplied.")
+
+    remove_unused_observables(problem)
+    remove_unused_conditions(problem)
+    condition_parameters_to_parameter_table(problem)
+
+    if lint_problem(problem):
+        raise AssertionError("Invalid PEtab problem generated.")
+
+
+def condition_parameters_to_parameter_table(problem: Problem):
+    """Move parameters from the condition table to the parameters table, if
+    the same parameter value is used for all conditions."""
+    if problem.condition_df is None or problem.condition_df.empty \
+            or problem.model is None:
+        return
+
+    replacements = {}
+    for parameter_id in problem.condition_df:
+        if parameter_id == PARAMETER_NAME:
+            continue
+
+        if problem.model.is_state_variable(parameter_id):
+            # initial states can't go the parameters table
+            continue
+
+        series = problem.condition_df[parameter_id]
+        if series.dtype == 'o':
+            # We don't touch parametric overrides
+            continue
+
+        # same value for all conditions?
+        if len(series.unique()) == 1:
+            replacements[parameter_id] = series[0]
+
+    if not replacements:
+        return
+
+    rows = [
+        {
+            PARAMETER_ID: parameter_id,
+            PARAMETER_SCALE: LIN,
+            LOWER_BOUND: nan,
+            UPPER_BOUND: nan,
+            NOMINAL_VALUE: value,
+            ESTIMATE: 0
+        }
+        for parameter_id, value in replacements.items()
+    ]
+    rows = pd.DataFrame(rows)
+    rows.set_index(PARAMETER_ID, inplace=True)
+
+    if problem.parameter_df is None:
+        problem.parameter_df = rows
+    else:
+        problem.parameter_df = pd.concat([problem.parameter_df, rows])
+
+    problem.condition_df = \
+        problem.condition_df.drop(columns=replacements.keys())
--- a/petab/simplify.py
+++ b/petab/simplify.py
@@ -3,6 +3,7 @@ from math import nan
 
 import pandas as pd
 
+import petab
 from . import Problem
 from .C import *  # noqa: F403
 from .lint import lint_problem
@@ -72,12 +73,10 @@ def condition_parameters_to_parameter_table(problem: Problem):
             continue
 
         series = problem.condition_df[parameter_id]
-        if series.dtype == 'o':
-            # We don't touch parametric overrides
-            continue
+        value = petab.to_float_if_float(series[0])
 
-        # same value for all conditions?
-        if len(series.unique()) == 1:
+        # same value for all conditions and no parametric overrides (str)?
+        if isinstance(value, float) and len(series.unique()) == 1:
             replacements[parameter_id] = series[0]
 
     if not replacements:
--- a/tests/test_simplify.py
+++ b/tests/test_simplify.py
@@ -57,6 +57,16 @@ def problem() -> Problem:
 
 
 def test_remove_nan_measurements(problem):
+    expected = pd.DataFrame(
+            {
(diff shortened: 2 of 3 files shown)


petab/simplify.py
```diff
@@ -0,0 +1,106 @@
...
+    "condition_parameters_to_parameter_table",
+]
+
+
+def remove_nan_measurements(problem: Problem):
+    """Drop any measurements that are NaN"""
+    problem.measurement_df = problem.measurement_df[
+        ~problem.measurement_df[MEASUREMENT].isna()
+    ]
+    problem.measurement_df.reset_index(inplace=True, drop=True)
```
dilpath, line 24: Why is this useful? Could be moved to `get_measurement_df` if useful

petab/simplify.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+def remove_unused_observables(problem: Problem):
+    """Remove observables that have no measurements"""
+    measured_observables = set(problem.measurement_df[OBSERVABLE_ID].unique())
```
dilpath, line 29: Should be equivalent
```suggestion
    measured_observables = set(problem.measurement_df[OBSERVABLE_ID])
```

petab/simplify.py
```diff
@@ -0,0 +1,106 @@
...
+    problem.observable_df = problem.observable_df[
+        problem.observable_df.index.isin(measured_observables)]
+
+
+def remove_unused_conditions(problem: Problem):
+    """Remove conditions that have no measurements"""
+    measured_conditions = \
+        set(problem.measurement_df[SIMULATION_CONDITION_ID].unique())
```
dilpath, line 37:
```suggestion
        set(problem.measurement_df[SIMULATION_CONDITION_ID])
```

petab/simplify.py
```diff
@@ -0,0 +1,106 @@
...
+    if PREEQUILIBRATION_CONDITION_ID in problem.measurement_df:
+        measured_conditions |= \
+            set(problem.measurement_df[PREEQUILIBRATION_CONDITION_ID].unique())
```
dilpath, line 40:
```suggestion
            set(problem.measurement_df[PREEQUILIBRATION_CONDITION_ID])
```

petab/simplify.py
```diff
@@ -0,0 +1,106 @@
...
+        problem.condition_df.index.isin(measured_conditions)]
+
+
+def simplify_problem(problem: Problem):
+    if lint_problem(problem):
+        raise ValueError("Invalid PEtab problem supplied.")
+
+    remove_unused_observables(problem)
+    remove_unused_conditions(problem)
+    condition_parameters_to_parameter_table(problem)
```
dilpath, line 52: Why is `remove_nan_measurements` missing here? Are `nan` measurements sometimes useful?

petab/simplify.py
```diff
@@ -0,0 +1,106 @@
...
+        set(problem.measurement_df[SIMULATION_CONDITION_ID].unique())
+    if PREEQUILIBRATION_CONDITION_ID in problem.measurement_df:
+        measured_conditions |= \
+            set(problem.measurement_df[PREEQUILIBRATION_CONDITION_ID].unique())
+
+    problem.condition_df = problem.condition_df[
+        problem.condition_df.index.isin(measured_conditions)]
+
+
+def simplify_problem(problem: Problem):
```
dilpath, line 46: Could add arguments to toggle each of the simplifications, and linting

petab/simplify.py
```diff
@@ -0,0 +1,106 @@
...
+def condition_parameters_to_parameter_table(problem: Problem):
+    """Move parameters from the condition table to the parameters table, if
+    the same parameter value is used for all conditions."""
+    if problem.condition_df is None or problem.condition_df.empty \
+            or problem.model is None:
+        return
+
+    replacements = {}
+    for parameter_id in problem.condition_df:
+        if parameter_id == PARAMETER_NAME:
```
dilpath, line 67:
```suggestion
        if parameter_id == CONDITION_NAME:
```

petab/simplify.py
```diff
@@ -0,0 +1,106 @@
...
+            continue
+
+        series = problem.condition_df[parameter_id]
+        if series.dtype == 'o':
+            # We don't touch parametric overrides
+            continue
+
+        # same value for all conditions?
+        if len(series.unique()) == 1:
+            replacements[parameter_id] = series[0]
```
dilpath, line 81: Is it always `'o'` for incompatible columns, e.g. not `str`? Could a column of floats be `'o'` anyway? A possible alternative.
```suggestion
        values = list(set(problem.condition_df[parameter_id]))
        if len(values) != 1:
            continue

        try:
            float_value = float(values[0])
            assert float_value == values[0]
        except:
            continue

        replacements[parameter_id] = float_value
```

tests/test_simplify.py
```diff
@@ -0,0 +1,149 @@
+"""Test for petab.simplify.*"""
```
dilpath, line 1:
```suggestion
"""Tests for petab.simplify.*"""
```

tests/test_simplify.py
```diff
@@ -0,0 +1,149 @@
...
+    ss_model.addParameter("same_value_for_all_conditions", val=1.0)
+
+    observable_df = pd.DataFrame(
+        {
+            OBSERVABLE_ID: ["obs_used", "obs_unused", "obs_used_2"],
+            OBSERVABLE_FORMULA: [1.0, 2.0, 3.0],
+            NOISE_FORMULA: [1.0, 2.0, 3.0],
+        }
+    )
+    observable_df.set_index(OBSERVABLE_ID, inplace=True)
```
dilpath, line 28: Could use `get_observable_df`

tests/test_simplify.py
```diff
@@ -0,0 +1,149 @@
...
+    conditions_df = pd.DataFrame(
+        {
+            CONDITION_ID: ["condition_used_1",
+                           "condition_unused",
+                           "condition_used_2"],
+            "some_parameter": [1.0, 2.0, 3.0],
+            "same_value_for_all_conditions": [4.0] * 3,
+        }
+    )
+    conditions_df.set_index(CONDITION_ID, inplace=True)
```
dilpath, line 39: Could use `get_condition_df`

tests/test_simplify.py
```diff
@@ -0,0 +1,149 @@
...
+            {
+                OBSERVABLE_ID: ["obs_used"] * 2,
+                MEASUREMENT: [1.0, 2.0],
+                SIMULATION_CONDITION_ID:
+                    ["condition_used_1", "condition_used_2"],
+                TIME: [1.0] * 2,
+            }
+        )
+    print(problem.measurement_df)
+    print(expected)
```
dilpath, line 83:
```suggestion
```

tests/test_simplify.py
```diff
@@ -0,0 +1,149 @@
...
+    assert_frame_equal(problem.measurement_df, expected)
```
dilpath, line 84: Could assert that they are not equal before applying the method, for this and other tests

tests/test_simplify.py
```diff
@@ -0,0 +1,149 @@
...
+                CONDITION_ID: ["condition_used_1",
+                               "condition_used_2"],
+                "some_parameter": [1.0, 3.0],
+                "same_value_for_all_conditions": [4.0] * 2,
+            }
+        )
+    expected.set_index(CONDITION_ID, inplace=True)
+
+    print(problem.condition_df)
+    print(expected)
```
dilpath, line 116:
```suggestion
```

tests/test_simplify.py
```diff
@@ -0,0 +1,149 @@
...
+    expected = pd.DataFrame(
+        {
+            CONDITION_ID: ["condition_used_1",
+                           "condition_unused",
+                           "condition_used_2"],
+            "some_parameter": [1.0, 2.0, 3.0],
+        }
+    )
+    expected.set_index(CONDITION_ID, inplace=True)
+    print(problem.condition_df, expected)
```
dilpath, line 132:
```suggestion
```

tests/test_simplify.py
```diff
@@ -0,0 +1,149 @@
...
+
+    expected = pd.DataFrame(
+        {
+            CONDITION_ID: ["condition_used_1",
+                           "condition_unused",
+                           "condition_used_2"],
+            "some_parameter": [1.0, 2.0, 3.0],
+        }
+    )
+    expected.set_index(CONDITION_ID, inplace=True)
```
dilpath, line 131: Could use `get_condition_df`

tests/test_simplify.py
```diff
@@ -0,0 +1,149 @@
...
+
+    expected = pd.DataFrame({
+            PARAMETER_ID: ["same_value_for_all_conditions"],
+            PARAMETER_SCALE: [LIN],
+            LOWER_BOUND: [nan],
+            UPPER_BOUND: [nan],
+            NOMINAL_VALUE: [4.0],
+            ESTIMATE: [0],
+        })
+    expected.set_index(PARAMETER_ID, inplace=True)
```
dilpath, line 143: Could use `get_condition_df`

dilpath (approved):
Looks good :+1: Fine as is except for the `PARAMETER_NAME` comment

petab/simplify.py
dweindl: Without that, you will have non-consecutive numbers in the index, which might be somewhat confusing. `get_measurement_df` wouldn't address that.

petab/simplify.py
dweindl: was hoping `unique` was happening in C and might be a bit faster.

petab/simplify.py
dweindl: as above

petab/simplify.py
dweindl: Because nan-measurements are not allowed (anymore). This would fail the initial validation step. I still added it, because it's useful for cleaning some older files, where it can be used outside this function.

petab/simplify.py
dweindl: Thought about it, but then again, just calling the functions you want individually takes about the same effort as specifying the options.

petab/simplify.py
dweindl: Ooops. :+1:

petab/simplify.py
dweindl: Yes, `str` would also be `O`. But you could convert an all-floats series to `O` too, right.

petab/simplify.py
dilpath: Ah, I meant if it's useful, then should `get_measurement_df` also do it?

dweindl: I don't know if somebody might prefer to keep the original index for whatever reason. For PEtab it wouldn't matter, it's not used anywhere. :man_shrugging:

tests/test_simplify.py
dweindl: I found it more explicit this way. Undecided.

petab/simplify.py
dweindl: Done

tests/test_simplify.py
dweindl: Done

PR #178, opened 2022-09-06
Add petab.Problem.__str__

dweindl:
... providing some high-level summary of the given problem.


petab/problem.py
```diff
@@ -104,6 +104,31 @@ def __setattr__(self, name, value):
...
+        observables = f"{self.observable_df.shape[0]} observables" \
+            if self.observable_df is not None else "without observables table"
+
+        measurements = f"{self.measurement_df.shape[0]} measurements" \
+            if self.measurement_df is not None \
+            else "without measurements table"
+
+        if self.parameter_df is not None:
+            num_estimated_parameters = sum(self.parameter_df[ESTIMATE] == 1) \
+                if ESTIMATE in self.parameter_df \
```
dilpath, line 121: `ESTIMATE` is a required column

petab/problem.py
```diff
@@ -104,6 +104,31 @@ def __setattr__(self, name, value):
         else:
             super().__setattr__(name, value)
 
+    def __str__(self):
+        model = f"with model ({self.model})" if self.model else "without model"
```
dilpath, line 108: `Model` has no `__str__`, what gets printed here?

dilpath (approved):
:+1:

petab/problem.py
dweindl: True. Not sure how much we want to assume/require correctness here. Should it just fail if this is missing? Should we validate in __str__? 
Might be more convenient to allow for some slack, if this is used during assembling a Problem. Similar to libsbml where one is allowed to create invalid models.

petab/problem.py
dweindl: Currently the not so informative default __repr__. The plan would be to add something more meaningful there as well.

PR #181, opened 2022-10-11
Update deprecated functions in tests

yannikschaelte:
I'd argue that when marking arguments as deprecated, tests should proceed to use the new arguments (and optionally keep some tests using the old ones, but only as extras)?


dweindl (approved):
> I'd argue that when marking arguments as deprecated, tests should proceed to use the new arguments (and optionally keep some tests using the old ones, but only as extras)?

Generally, I'd say it makes sense to check that the deprecated functionality still works until it is removed. But fine for me to remove in this specific case.

I added test routines with deprecation warning checks for most of the changed calls

PR #185, opened 2022-12-06
Vis: Don't fail on missing simulations

dweindl:


petab/visualize/plotter.py
```diff
@@ -669,7 +670,7 @@ def check_data_to_plot(
             """
             contains_inf = False
             max_finite_cond, min_cond = None, np.inf
-            if data_to_plot is not None:
+            if data_to_plot is not None and len(data_to_plot.conditions):
```
dilpath, line 673: Data is always associated with a condition, right? Why might there be data and no conditions?

dilpath (approved):

petab/visualize/plotter.py
dweindl: > Data is always associated with a condition, right? 

Right. 

> Why might there be data and no conditions?

In the current implementation, `data_to_plot` can be empty, i.e. not contain any datapoints / conditions (without being `None`). 
Overall, this change makes it easier to share a visualization file between different problems. If some measurements are missing, this will just create an empty plot instead of raising some non-informative exception.

PR #191, opened 2023-01-31
add histogram to residuals plot and center zero

plakrisenko:


petab/visualize/plot_residuals.py
```diff
@@ -13,17 +13,16 @@
 from ..calculate import calculate_residuals
 from ..core import get_simulation_df
 from ..problem import Problem
-from ..C import NOISE_DISTRIBUTION, NORMAL, LIN, OBSERVABLE_TRANSFORMATION, \
-    OBSERVABLE_ID
+from ..C import *
```
PaulJonasJost, line 16: is this actually clean? would think it would be cleaner to still import only the ones used.

PaulJonasJost (approved):

PR #192, opened 2023-01-31
add goodness of fit plot

plakrisenko:


petab/visualize/plot_residuals.py
```diff
@@ -100,3 +101,81 @@ def plot_residuals_vs_simulation(
...
+                                  'are needed for goodness_of_fit')
+
+    residual_df = calculate_residuals(
+        measurement_dfs=petab_problem.measurement_df,
+        simulation_dfs=simulations_df,
+        observable_dfs=petab_problem.observable_df,
+        parameter_dfs=petab_problem.parameter_df)[0]
+    slope, intercept, r_value, p_value, std_err = stats.linregress(
+        petab_problem.measurement_df['measurement'],
+        simulations_df['simulation'])  # x, y
```
PaulJonasJost, line 146: you calculate here the linear regression between measurement and simulation. What would  it tell me if this is not a diagonal plot?

PaulJonasJost (approved):
Seems like a useful verification of a good fit. Not sure on specific measures in case it is not a good fit.

dilpath (approved):
:+1:

Could add an example to a notebook

petab/visualize/plot_residuals.py
plakrisenko: You could see here if simulated values tend to be larger or smaller that measurement values. This plot probably won't be useful in every case. Also it gives some measure of how much off the simulation is.  
It could be useful if one wants to compare different approaches to fit the model.

PR #193, opened 2023-02-13
prevent strings being parsed as nan in get_visualization_df

plakrisenko:


petab/core.py
```diff
@@ -65,7 +65,8 @@ def get_visualization_df(visualization_file: Union[str, Path]) -> pd.DataFrame:
     """
     try:
         vis_spec = pd.read_csv(visualization_file, sep="\t", index_col=None,
-                               float_precision='round_trip')
+                               float_precision='round_trip',
+                               keep_default_na=False)
```
dweindl, line 69: I think, I would prefer going via `dtype` to avoid potential issues elsewhere. That should also solve the issue, right?

plakrisenko: dtype didn't help, but converters did

dweindl (approved):
:+1:

PR #194, opened 2023-02-20
Fix get_model_for_condition

dweindl:
Handle parametric initial concentrations for cases where the initial concentration parameter does not occur in the model, but only in the parameter table.


dilpath (approved):
:+1:

PR #196, opened 2023-03-09
Add pyproject.toml

dweindl:
Fixes:

>  DEPRECATION: petab is being installed using the legacy 'setup.py install' method, because it does not have a 'pyproject.toml' and the 'wheel' package is not installed. pip 23.1 will enforce this behaviour change. A possible replacement is to enable the '--use-pep517' option. Discussion can be found at https://github.com/pypa/pip/issues/8559

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,6 @@
+[build-system]
+requires = [
+    "setuptools>=61",
+    "wheel",
+]
+build-backend = "setuptools.build_meta"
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -0,0 +1,6 @@
+[build-system]
+requires = [
+    "setuptools>=61",
```
dilpath, line 3:
```suggestion
    "setuptools>=62",
```

for this, but fine as is for me: https://github.com/pypa/setuptools/pull/3151

dilpath (approved):

PR #199, opened 2023-04-22
Simulator: rename measurement column to simulation

dilpath:
What's your opinion on whether to default to `measurement` or `simulation`? Either is fine for me. Currently I am in favor of keeping `measurement`, because:
- I have mostly used this for synthetic data generation
- It won't break user code

However, I see that `simulation` might be what users expect from a simulator.

see also: AMICI-dev/AMICI#2050 and AMICI-dev/AMICI#2051

### DIFF
--- a/petab/simulate.py
+++ b/petab/simulate.py
@@ -137,10 +137,12 @@ class Simulator(abc.ABC):
         simulation_df = self.simulate_without_noise(**kwargs)
         if noise:
             simulation_df = self.add_noise(simulation_df, noise_scaling_factor)
-        if not as_measurement:
-            simulation_df = simulation_df.rename(
-                columns={petab.C.MEASUREMENT: petab.C.SIMULATION},
-            )
+
+        columns = {petab.C.MEASUREMENT: petab.C.SIMULATION}
+        if as_measurement:
+            columns = {petab.C.SIMULATION: petab.C.MEASUREMENT}
+        simulation_df = simulation_df.rename(columns=columns)
+
         return simulation_df
 
     def add_noise(
--- a/petab/simulate.py
+++ b/petab/simulate.py
@@ -115,7 +115,7 @@ class Simulator(abc.ABC):
             self,
             noise: bool = False,
             noise_scaling_factor: float = 1,
-            rename: bool = False,
+            as_measurement: bool = False,
             **kwargs
     ) -> pd.DataFrame:
         """Simulate a PEtab problem, optionally with noise.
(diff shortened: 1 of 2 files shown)


dweindl:
> However, I see that `simulation` might be what users expect from a simulator.

That. 

>     * I have mostly used this for synthetic data generation

Maybe we should split it into a Simulator and a SyntheticDataGenerator, each with their natural default?

>     * It won't break user code

Fair point, but I'd say it is not so widely used yet, that I'd rather avoid confusing future users.



Would change `rename` to `as_measurement` or `as_simulation`, depending on the final choice of the default.

fbergmann:
Looks good to me, though i of course would prefer to have the default for rename set to `True`, since I'd mainly expect a simulation dataframe when calling simulate.

petab/simulate.py
```diff
@@ -115,6 +115,7 @@ def simulate(
             self,
             noise: bool = False,
             noise_scaling_factor: float = 1,
+            rename: bool = False,
```
fbergmann, line 118: would you consider changing this to `True`?

dilpath:
Thanks for the feedback! I opted to default to `as_measurement=False`, and otherwise keep it as-is.

dweindl (approved):
Would be good to specify/document what `simulate_without_noise` is supposed to return - `measurement` or `simulation`.
To me it would feel most natural to have `simulation` for no-noise, and measurement for with-noise. I don't know if this might be considered confusing by others.

dilpath:
> Would be good to specify/document what `simulate_without_noise` is supposed to return - `measurement` or `simulation`.

Hm, at the moment the PEtab spec states
> (optional) A simulation file, which has the same format as the measurement file, but contains model simulations [TSV]

But this looks like a typo, since the `simulation` column name is used in practice.

Currently, the docstring for `simulate_without_noise` specifies `measurement` (as in the spec):

> Simulated data, as a PEtab measurements table, which should be equivalent to replacing all values in the :const:`petab.C.MEASUREMENT` column of the measurements table (of the PEtab problem supplied to the :meth:`petab.simulate.Simulator.__init__` method), with simulated values.

But the `simulate` method now enforces a specific column name, so the user always sees consistent column names.

> To me it would feel most natural to have `simulation` for no-noise, and `measurement` for with-noise. I don't know if this might be considered confusing by others.

Currently, as user can choose the column name, with `as_measurement`. Are you suggesting that I remove `as_measurement` and instead use `Simulator.simulate(noise=True/False)` determine the column name? The user is never meant to use `simulate_without_noise`, rather use `Simulator.simulate(noise=True/False)`.

dweindl:
> > Would be good to specify/document what `simulate_without_noise` is supposed to return - `measurement` or `simulation`.
> 
> Hm, at the moment the PEtab spec states
> 
> > (optional) A simulation file, which has the same format as the measurement file, but contains model simulations [TSV]
> 
> But this looks like a typo, since the `simulation` column name is used in practice.

Agreed, should be changed.

> Currently, the docstring for `simulate_without_noise` specifies `measurement` (as in the spec):
> 
> > Simulated data, as a PEtab measurements table, which should be equivalent to replacing all values in the :const:`petab.C.MEASUREMENT` column of the measurements table (of the PEtab problem supplied to the :meth:`petab.simulate.Simulator.__init__` method), with simulated values.

Sorry, I missed that. I accidentally looked at `TestSimulator`, which didn't have an informative docstring. All good.

> > To me it would feel most natural to have `simulation` for no-noise, and `measurement` for with-noise. I don't know if this might be considered confusing by others.
> 
> Currently, as user can choose the column name, with `as_measurement`. Are you suggesting that I remove `as_measurement` and instead use `Simulator.simulate(noise=True/False)` determine the column name? The user is never meant to use `simulate_without_noise`, rather use `Simulator.simulate(noise=True/False)`.

No, I think it's clearly specified then.

PR #200, opened 2023-05-05
Require Python 3.9

dweindl:


setup.py
```diff
@@ -29,8 +29,8 @@ def absolute_links(txt):
 
 
 # Python version check. We need >= 3.6 due to e.g. f-strings
```
fbergmann, line 31: better update the comment too! .. but really 3.9 already?

dweindl: Agreed. Since we rely on numpy, I think it makes sense to follow https://numpy.org/neps/nep-0029-deprecation_policy.html, which drops Python 3.8 support on April 14. Not sure if this was officially discussed.

fbergmann: might be something to think about for those poor people still on older versions desperate to support petab. I think the numpy deprecation would be fine if libpetab depended on any of the C API. But since that is not the case, there is literally nothing in libpetab that would break with an older bumpy version.

dweindl: I don't think this is only a C API/ABI issue. See e.g. https://numpy.org/doc/stable/release/1.24.0-notes.html#expired-deprecations

fbergmann (approved):
alright all good with me ... those poor people with py38 will have the last release of petab to be happy with

fbergmann:
OMG ... readthedocs does not do py39 yet ... 

rtd

told you forcing it to 3.9 might not be a good idea

dweindl:
It does, if we ask nicely. Wait for it...

PR #202, opened 2023-05-08
Fix sympy symbol name clashes

dweindl:
Make sure symbols like `N`, `beta`, ... are used as scalar symbols, and not as sympy functions with the same name.

See also https://github.com/ICB-DCM/pyPESTO/issues/1048


dilpath (approved):
:+1:

PR #208, opened 2023-06-20
GHA/pypi: trusted publishing

dweindl:
No need for long-lived API tokens anymore. see https://docs.pypi.org/trusted-publishers/


dilpath (approved):
:+1:

PR #210, opened 2023-06-21
Filter/expect/prevent warnings in test suite 

dweindl:
Test suite now completes without any distracting warnings.


dilpath (approved):
Nice :+1:

PR #228, opened 2023-09-19
Fix accessing preequilibrationConditionId without checking for presence

dweindl:
Fixes a bug in petab.visualize.data_overview.create_report which fails if there is no preequilibrationConditionId column in the measurements table.

### DIFF
--- a/petab/visualize/data_overview.py
+++ b/petab/visualize/data_overview.py
@@ -61,13 +61,14 @@ def get_data_per_observable(measurement_df: pd.DataFrame) -> pd.DataFrame:
     """
 
     my_measurements = measurement_df.copy()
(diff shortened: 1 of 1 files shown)


dilpath (approved):
:+1:

The old code assumes the existence of `my_measurements[PREEQUILIBRATION_CONDITION_ID]`, so maybe this assumption is made elsewhere too and it's safer to create the `PREEQUILIBRATION_CONDITION_ID` key and fill it with dummy values here if it doesn't exist. I don't really know though, fine as is for me.

dweindl:
> safer to create the `PREEQUILIBRATION_CONDITION_ID` key and fill it with dummy values here if it doesn't exist

Yes. That would be something to consider throughout the library. Has been discussed before, but without any clear conclusion. I will leave it as is for now.

PR #230, opened 2023-11-01
Startpoint sampling for a subset of parameters

dweindl:
Allow passing a list of parameter IDs to startpoint sampling for subsetting/reordering parameters.

### DIFF
--- a/petab/parameters.py
+++ b/petab/parameters.py
@@ -424,19 +424,25 @@ def get_valid_parameters_for_parameter_table(
 def get_priors_from_df(
     parameter_df: pd.DataFrame,
     mode: Literal["initialization", "objective"],
+    parameter_ids: Sequence[str] = None,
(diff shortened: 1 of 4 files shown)


plakrisenko (approved):

petab/parameters.py
```diff
@@ -424,20 +424,26 @@ def append_overrides(overrides):
...
 
     Returns:
         List with prior information.
     """
-
     # get types and parameters of priors from dataframe
     par_to_estimate = parameter_df.loc[parameter_df[ESTIMATE] == 1]
 
+    if parameter_ids:
+        par_to_estimate = par_to_estimate.loc[parameter_ids, :]
```
plakrisenko, line 445: Maybe add an error message for the case when provided parameter_ids list contains an id that is not there.

PR #231, opened 2023-12-11
Treat observableParameter overrides as placeholders in noise formulae

dilpath:
Currently, linting will fail if an `observableParameter` (e.g. a scaling or offset) appears in the noise formulae. Having the full observable formulae in the noise formulae is currently the only way to get observable-dependent noise -- this means noise formulae should support `observableParameter` placeholders.

This fixes the issue for this case: https://github.com/Benchmarking-Initiative/Benchmark-Models-PEtab/pull/197

If I compare the AMICI model from these cases:
- observable-dependent noise via observable ID in noise formulae (invalid PEtab, but compiles fine with AMICI)
- this fix + observable-dependent noise via observable formulae in the noise formulae
I get the same C++ files.

### DIFF
--- a/petab/parameters.py
+++ b/petab/parameters.py
@@ -300,7 +300,7 @@ def get_required_parameters_for_parameter_table(
     # Add output parameters except for placeholders
     for kwargs in [
         dict(observables=True, noise=False),
-        dict(observables=False, noise=True),
+        dict(observables=True, noise=True),
     ]:
         output_parameters = observables.get_output_parameters(
             observable_df, model, mapping_df=mapping_df, **kwargs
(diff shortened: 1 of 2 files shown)


dweindl (approved):
Thanks. Makes sense. Could you please add/extend a test?

We should probably also clarify that in https://github.com/PEtab-dev/PEtab/blob/main/doc/documentation_data_format.rst

dilpath:
> Thanks. Makes sense. Could you please add/extend a test?

Done! I did not see a pre-existing test, so I created a new one in a test script that already provided a PEtab problem to work with. I changed this PEtab problem, because `observable_1` is a parameter in the test problem model, and `obs1` is the observable ID used in the measurements table.

> We should probably also clarify that in https://github.com/PEtab-dev/PEtab/blob/main/doc/documentation_data_format.rst

Sounds good, done here: https://github.com/PEtab-dev/PEtab/pull/574

PR #239, opened 2023-12-18
Fix flatten_timepoint_specific_output_overrides

dweindl:
... again.

Fixes an error when trying to apply re.sub to numeric values:

```
    petab.flatten_timepoint_specific_output_overrides(problem)
build/venv/lib/python3.9/site-packages/petab/core.py:290: in flatten_timepoint_specific_output_overrides
    observable[target] = re.sub(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

pattern = 'observableParameter([0-9]+)_obs_a'
repl = 'observableParameter\\1_obs_a__10__c0', string = 1, count = 0, flags = 0

    def sub(pattern, repl, string, count=0, flags=0):
        """Return the string obtained by replacing the leftmost
        non-overlapping occurrences of the pattern in string by the
        replacement repl.  repl can be either a string or a callable;
        if a string, backslash escapes in it are processed.  If it is
        a callable, it's passed the Match object and must return
        a replacement string to be used."""
>       return _compile(pattern, flags).sub(repl, string, count)
E       TypeError: cannot use a string pattern on a bytes-like object

/opt/hostedtoolcache/Python/3.9.18/x64/lib/python3.9/re.py:210: TypeError
```

### DIFF
--- a/petab/core.py
+++ b/petab/core.py
@@ -17,6 +17,7 @@ from warnings import warn
 
 import numpy as np
 import pandas as pd
+from pandas.api.types import is_string_dtype
 
 from . import yaml
 from .C import *  # noqa: F403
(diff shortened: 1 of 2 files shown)


dilpath (approved):
:+1:

PR #243, opened 2024-01-08
Add petab.Problem.n_{estimated,measurements,priors}

dweindl:
Makes it more convenient to compute model selection criteria.


dilpath (approved):
:+1:

PR #248, opened 2024-03-05
Fix `SbmlModel.get_free_parameter_ids_with_values` to handle InitialA…

dweindl:
…ssignments

So far, `SbmlModel.get_free_parameter_ids_with_values` ignored InitialAssignments, resulting in incorrect parameter values in the parameter mapping.

### DIFF
--- a/petab/models/sbml_model.py
+++ b/petab/models/sbml_model.py
@@ -5,6 +5,7 @@ from pathlib import Path
 from typing import Iterable, Optional, Tuple
 
 import libsbml
+import sympy as sp
 
 from ..sbml import (
     get_sbml_model,
@@ -103,8 +104,31 @@ class SbmlModel(Model):
             ar.getVariable() for ar in self.sbml_model.getListOfRules()
         }
 
+        parser_settings = libsbml.L3ParserSettings(
+            self.sbml_model,
+            libsbml.L3P_PARSE_LOG_AS_LOG10,
+            libsbml.L3P_EXPAND_UNARY_MINUS,
+            libsbml.L3P_NO_UNITS,
+            libsbml.L3P_AVOGADRO_IS_CSYMBOL,
+            libsbml.L3P_COMPARE_BUILTINS_CASE_INSENSITIVE,
+            None,
+            libsbml.L3P_MODULO_IS_PIECEWISE,
+        )
+
+        def get_initial(p):
+            # return the initial assignment value if there is one, and it is a
+            # number, otherwise the parameter value
+            if ia := self.sbml_model.getInitialAssignmentBySymbol(p.getId()):
+                sym_expr = sp.sympify(
+                    libsbml.formulaToL3StringWithSettings(
+                        ia.getMath(), parser_settings
+                    )
+                )
+                return float(sym_expr) if sym_expr.is_Number else p.getValue()
+            return p.getValue()
+
         return (
-            (p.getId(), p.getValue())
+            (p.getId(), get_initial(p))
             for p in self.sbml_model.getListOfParameters()
             if p.getId() not in rule_targets
         )
--- a/petab/models/sbml_model.py
+++ b/petab/models/sbml_model.py
@@ -117,20 +117,28 @@ class SbmlModel(Model):
 
         def get_initial(p):
             # return the initial assignment value if there is one, and it is a
-            # number, otherwise the parameter value
+            # number; `None`, if there is a non-numeric initial assignment;
+            # otherwise, the parameter value
             if ia := self.sbml_model.getInitialAssignmentBySymbol(p.getId()):
-                sym_expr = sp.sympify(
-                    libsbml.formulaToL3StringWithSettings(
-                        ia.getMath(), parser_settings
-                    )
+                formula_str = libsbml.formulaToL3StringWithSettings(
+                    ia.getMath(), parser_settings
                 )
-                return float(sym_expr) if sym_expr.is_Number else p.getValue()
+                try:
+                    return float(formula_str)
+                except ValueError:
+                    sym_expr = sp.sympify(formula_str)
+                    return (
+                        float(sym_expr.evalf())
+                        if sym_expr.evalf().is_Number
+                        else None
+                    )
             return p.getValue()
 
         return (
-            (p.getId(), get_initial(p))
+            (p.getId(), initial)
             for p in self.sbml_model.getListOfParameters()
             if p.getId() not in rule_targets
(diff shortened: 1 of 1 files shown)


petab/models/sbml_model.py
```diff
@@ -103,8 +104,31 @@ def get_free_parameter_ids_with_values(
...
+            libsbml.L3P_AVOGADRO_IS_CSYMBOL,
+            libsbml.L3P_COMPARE_BUILTINS_CASE_INSENSITIVE,
+            None,
+            libsbml.L3P_MODULO_IS_PIECEWISE,
+        )
+
+        def get_initial(p):
+            # return the initial assignment value if there is one, and it is a
+            # number, otherwise the parameter value
+            if ia := self.sbml_model.getInitialAssignmentBySymbol(p.getId()):
```
dilpath, line 121:
```suggestion
            if ia := self.sbml_model.getInitialAssignmentBySymbol(p.getId()) is not None:
```

petab/models/sbml_model.py
```diff
@@ -103,8 +104,31 @@ def get_free_parameter_ids_with_values(
             ar.getVariable() for ar in self.sbml_model.getListOfRules()
         }
 
+        parser_settings = libsbml.L3ParserSettings(
+            self.sbml_model,
+            libsbml.L3P_PARSE_LOG_AS_LOG10,
```
dilpath, line 109: This is pretty unintuitive, to me it looks like this will convert `log` to `log10`, and I'm not sure if the libsbml docs make sense. e.g. this for `L3P_PARSE_LOG_AS_ERROR`
> Collapse unary minuses where possible when parsing text-string formulas.

https://synonym.caltech.edu/software/libsbml/5.18.0/docs/formatted/python-api/namespacelibsbml.html#a9ef0398726fb45bd1cb24c4a5ed671f0

The docstring for `L3P_PARSE_LOG_AS_LOG10` looks like it actually parses `log` as `log`, not parsing `log` as `log10`, but I'm not sure I trust the docs given the docstring for `L3P_PARSE_LOG_AS_ERROR`. Should we instead use `L3P_PARSE_LOG_AS_LN`?

I think most users would expect `log` to be `ln`, and that `log10` is only when explicitly specified with `log(10, x)` or `log10(x)`.

dilpath (approved):
Why does a free parameter take a value from an SBML assignment rule? Shouldn't values for free parameters come from an optimizer? I would expect that initial assignments/values for free parameters are somehow removed, is that what this method is for?

petab/models/sbml_model.py
dweindl: That's implied.

petab/models/sbml_model.py
dweindl: Those are the libsbml default settings. The only relevant thing here is ignoring units.

dweindl:
> Why does a free parameter take a value from an SBML assignment rule? Shouldn't values for free parameters come from an optimizer? I would expect that initial assignments/values for free parameters are somehow removed, is that what this method is for?

Assignment rule targets are excluded here. 

The method name is maybe not ideal. In the end, this method is used to collect parameters and default values that should be present in the parameter mapping. I guess it's up for discussion what should be included/excluded there. However, the original situation was clearly bad: For example, a parameter with `value=0` and initialAssignment of `2` would have been included there with a value of `0` instead of `2`. For amici that didn't matter until recently (https://github.com/AMICI-dev/AMICI/pull/2304) due to its handling of initialAssignments, but for any other tool this would have been a problem.

petab/models/sbml_model.py
dilpath: OK, if it actually keeps `log` as `log`, not `log10`, then no problem

petab/models/sbml_model.py
dilpath: :+1: was just a PEP 8 recommendation

dilpath:
> > Why does a free parameter take a value from an SBML assignment rule? Shouldn't values for free parameters come from an optimizer? I would expect that initial assignments/values for free parameters are somehow removed, is that what this method is for?
> 
> Assignment rule targets are excluded here.

Meant to say initial assignment rule there, but I see now that this is not about free parameters only

> The method name is maybe not ideal. In the end, this method is used to collect parameters and default values that should be present in the parameter mapping. I guess it's up for discussion what should be included/excluded there. However, the original situation was clearly bad: For example, a parameter with `value=0` and initialAssignment of `2` would have been included there with a value of `0` instead of `2`. For amici that didn't matter until recently ([AMICI-dev/AMICI#2304](https://github.com/AMICI-dev/AMICI/pull/2304)) due to its handling of initialAssignments, but for any other tool this would have been a problem. Whether or not initialAssignments that depend on other parameters should be included here is a different question. Not sure about that. Including them, would however require more widespread changes, since then the parameter mapping would contain symbolic expressions. (Which would happen anyways if we proceed with the timecourse proposal).

OK, sounds good to me then. Doing SymPy calls frequently sounds inefficient, but maybe JIT/`eval`/`exec` can be used to turn them into a standard Python function on the fly.

petab/models/sbml_model.py
dweindl: Note that this is about parsing MathML not, any observableFormula or the like. In MathML `log` without explicit base [means `log10`](https://www.w3.org/TR/mathml4/#contm_log_sec).

dweindl:
> Doing SymPy calls frequently sounds inefficient, but maybe JIT/`eval`/`exec` can be used to turn them into a standard Python function on the fly.

This is happening once during the parameter mapping. I don't think it really matters, unless the model is really huge, and even then, this is probably not the major problem. I can try `float(.)` first to handle the super basic cases first without sympy, but I think it won't matter mach.

petab/models/sbml_model.py
dilpath: Oh! Good to know that `log` means different things in PEtab and the model :grimacing: Then I think we could be more careful about a math spec in the next version of PEtab, I added a comment to the draft timecourse spec.

dilpath:
> > Doing SymPy calls frequently sounds inefficient, but maybe JIT/`eval`/`exec` can be used to turn them into a standard Python function on the fly.
> 
> This is happening once during the parameter mapping. I don't think it really matters, unless the model is really huge, and even then, this is probably not the major problem. I can try `float(.)` first to handle the super basic cases first without sympy, but I think it won't matter mach.

OK, sounds reasonable to me; only initial assignments without parameter dependencies are handled.

PR #250, opened 2024-03-12
Typehint states it is optional, but no default value is provided.

PaulJonasJost:
Add None as default value, due to optionality.

### DIFF
--- a/petab/parameter_mapping.py
+++ b/petab/parameter_mapping.py
@@ -307,7 +307,7 @@ def _map_condition(packed_args):
 def get_parameter_mapping_for_condition(
     condition_id: str,
     is_preeq: bool,
-    cur_measurement_df: Optional[pd.DataFrame],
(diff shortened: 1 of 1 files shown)


dweindl (approved):
> Typehint states it is optional, but no default value is provided

`x: Optional[pd.DataFrame]` does not imply that it's an optional argument (i.e. should have a default value). It just means that `None` is a valid value. The naming is very unfortunate. But fine to default to `None` in this specific case.

PaulJonasJost:
> > Typehint states it is optional, but no default value is provided
> 
> `x: Optional[pd.DataFrame]` does not imply that it's an optional argument (i.e. should have a default value). It just means that `None` is a valid value. The naming is very unfortunate. But fine to default to `None` in this specific case.

True perhaps the naming is just unfortunate. Also saw no problem in None as default value. Thanks, will merge.

PR #251, opened 2024-04-16
maint: black -> ruff

dweindl:
Instead of black and isort, run ruff as pre-commit hook. Enable pyupgrade and additional rules.

Applied safe auto-fixes, but validation currently fails. Will be addressed in a follow-up PR (#252).

Relevant changes: `.pre-commit-config.yaml`, `pyproject.toml`


dilpath (approved):
:rocket:

PR #254, opened 2024-04-16
Vis: Make line plot markers configurable

dweindl:
Allow changing the default line style and markers for line plots.

(I'd also prefer to change the default to points-only (no line) for the measurements, but that's up for discussion.)

### DIFF
--- a/petab/visualize/plotter.py
+++ b/petab/visualize/plotter.py
@@ -15,14 +15,21 @@ from .plotting import DataPlot, DataProvider, DataSeries, Figure, Subplot
 
 __all__ = ["Plotter", "MPLPlotter", "SeabornPlotter"]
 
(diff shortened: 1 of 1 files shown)


dilpath (approved):
Fine for me as-is. Alternatively, make it a `dict` like
```python
petab.visualize.plotter.measurement_style = {
    linestyle: "-.",
    marker: "x",
}
```
then parse plot-specific options like
```python
plot_kwargs = plot_specific_defaults | petab.visualize.plotter.measurement_style
```

since then users can specify more plot options like `markersize`.

dilpath:
> (I'd also prefer to change the default to points-only (no line) for the measurements, but that's up for discussion.)

Also my preference :+1:

PR #259, opened 2024-05-14
Require positive bounds for log-scale parameters

dweindl:
Previously, bounds were only checked for non-negativity.

### DIFF
--- a/petab/lint.py
+++ b/petab/lint.py
@@ -547,7 +547,7 @@ def check_parameter_bounds(parameter_df: pd.DataFrame) -> None:
                     f"{LOWER_BOUND} greater than {UPPER_BOUND} for "
                     f"{PARAMETER_ID} {row.name}."
                 )
-            if (row[LOWER_BOUND] < 0.0 or row[UPPER_BOUND] < 0.0) and row[
+            if (row[LOWER_BOUND] <= 0.0 or row[UPPER_BOUND] <= 0.0) and row[
                 PARAMETER_SCALE
             ] in [LOG, LOG10]:
                 raise AssertionError(


MerktSimon (approved):

dilpath (approved):
Fine for me but it means tools cannot implement model selection via regularization with L1 (admittedly nontrivial on log-scale), because that would contradict the nonzero lower bound.

dweindl:
> Fine for me but it means tools cannot implement model selection via regularization with L1 (admittedly nontrivial on log-scale), because that would contradict the nonzero lower bound.

Thx. No strong opinion. I'm fine with either:
* allowing 0, but making that explicit in the specs and silencing all warnings related to log(0) in the library
* excluding 0

Other opinions?

dilpath:
Also no strong opinion.

We support nominal values of 0 regardless of parameter scale and bounds, so no issue for PEtab Select.

Since excluding 0 would be a possibly-breaking change, I would be in favor of allowing 0.

dweindl:
> Since excluding 0 would be a possibly-breaking change, I would be in favor of allowing 0.

Okay, I guess we'll leave it as is for PEtab v1.

For PEtab v2 we'll probably allow 0. For further discussion -> https://github.com/PEtab-dev/PEtab/pull/579

dilpath:
Sampling, e.g. initial start vectors for multi-start optimization, when the lower bound is `0` and the parameter is `log10`-scaled, would mean sampling in `(-inf, upperBound]`, which isn't sensible.

We could require that, if a user specifies a `lowerBound` of `0` for a `log10`-scaled parameter, then they should also specify `initializationPriorParameters` such that the initial sampling of start vectors can occur on `parameterScale` within finite bounds.

A PEtab problem where samples are drawn uniformly from non-finite bounds, should be considered invalid.

from discussion with @dweindl

dweindl:
Closing in favor of https://github.com/PEtab-dev/libpetab-python/pull/278

PR #260, opened 2024-05-21
Add PEtab math parser and sympy converter

dweindl:
Related to https://github.com/PEtab-dev/PEtab/pull/579

* ANTLR grammar for PEtab math expressions + derived lexer/parser/visitor
* Test cases
* PEtab -> SymPy converter

Note: All sympy symbols in PEtab match expressions now have `real=True`.

PR #261, opened 2024-05-21
Refactor sympification

dweindl:
Centralize sympification of PEtab and SBML math expressions.

Simplifies incorporating https://github.com/PEtab-dev/libpetab-python/pull/260

Also fixes likely collisions of SBML model entities with sympy entities.

### DIFF
--- a/petab/calculate.py
+++ b/petab/calculate.py
@@ -7,11 +7,11 @@ from typing import Dict, List, Union
 import numpy as np
 import pandas as pd
 import sympy
-from sympy.abc import _clash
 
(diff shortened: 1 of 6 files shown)


dilpath (approved):
What will the benefit of having `math.py` in its own directory be? For model-format--specific math methods?

dweindl:
> What will the benefit of having `math.py` in its own directory be? For model-format--specific math methods?

I am expecting more files to be added there: https://github.com/PEtab-dev/libpetab-python/pull/260

PR #264, opened 2024-05-30
Simplify yaml schema

dweindl:
Avoid repetitions in yaml schema.

### DIFF
--- a/petab/schemas/petab_schema.v2.0.0.yaml
+++ b/petab/schemas/petab_schema.v2.0.0.yaml
@@ -1,4 +1,4 @@
-# For syntax see: https://json-schema.org/understanding-json-schema/index.html
+# For syntax see: https://json-schema.org/understanding-json-schema
 #$schema: "https://json-schema.org/draft/2019-09/meta/core"
 $schema: "http://json-schema.org/draft-06/schema"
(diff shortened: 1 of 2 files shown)


petab/schemas/petab_schema.v2.0.0.yaml
dilpath, line 1: Top of this file: the URL `https://json-schema.org/understanding-json-schema/index.html` could be changed to `https://json-schema.org/understanding-json-schema`.

Also confusing to read this is a JSON schema, but it's in YAML. Does the JSON schema cover this YAML document explicitly? Or is it just that this document describes some dictionary that JSON schema is compatible with, but the YAML directly is not supported?

dilpath (approved):
:+1:

petab/schemas/petab_schema.v2.0.0.yaml
dweindl: json-schema has been developed for json, but it can be applied to Python objects. We are applying it to a `dict`. It does not check the yaml syntax itself (loading the yaml file will already fail before the validation in case of syntax errors), but only the data model.

dweindl:
- [ ] the v1 schema should also be updated

Issue #271, opened 2024-06-25
Library organization for petab v2


dweindl:
We are planning for a PEtab v2 release, ideally within the next 6 months. An open question is how to handle that in libpetab, as there will be quite some incompatible changes. This issue is to discuss/outline how we accommodate these changes here.

The goal should be to support PEtab v1 and PEtab v2 files in parallel. We should be able to validate both v1 and v2 files and there should be some functionality to up-convert PEtab v1 problems.

My current idea would be to have `petab.v1` and  `petab.v2` subpackages, that can be imported with `import petab.v1 as petab` and be used without (or very few) adaptions. This will imply some code duplication, but I still think it will be more maintainable.

I would go for a libpetab 1.0 release where we clarify the public API and move code to a `petab.v1 subpackage`.
I am not sure if we want to keep things importable from `petab` directly. I tend to require `import petab.v1 as petab`.

dilpath:
The suggestion of `petab.v1/v2` sounds best to me. And I would also be in favor for requiring `import petab.v1 as petab`. Is a deprecation period possible for this major change? We could add `from .v1 import *` to `petab/__init__.py` I guess.

Things like `petablint` will require version information as well, unless the problem YAML with version information is supplied. Or would we also rename that to `petablint.v1/v2`?

dweindl:
> Is a deprecation period possible for this major change? We could add `from .v1 import *` to `petab/__init__.py` I guess.

Yes. And introduce `petab.v1` rather sooner than later, so that consumers can already adapt to that.

> Things like `petablint` will require version information as well, unless the problem YAML with version information is supplied. Or would we also rename that to `petablint.v1/v2`?

I think the main use case is running `petablint` on a full petab problem where the version number would be available from the yaml file. I am not sure how many people are running it on individual files where validation is anyways rather limited.
Either we add a version option for validation of individual files, or – preferred – only handle full petab problems (and make the `-y` optional and remove it at some point).

dilpath:
>  Either we add a version option for validation of individual files, or – preferred – only handle full petab problems (and make the -y optional and remove it at some point).

Makes sense, then we could deprecate validation of individual tables already, with some message asking for feedback if someone needs it.

dweindl:
> we could deprecate validation of individual tables already

https://github.com/PEtab-dev/libpetab-python/pull/274

Please post any feedback specific to the adaptation of `petablint` there.

PR #274, opened 2024-06-25
Deprecate petablint with individual tables

dweindl:
I think the main use case is running petablint on a full petab problem, so we don't need the option to pass individual tables. To support individual tables with the upcoming petab v2, we'd have to extend the list of files and require a version number. I don't it's worth maintaining this functionality (correct me if I am wrong), since validation of individual files is anyways rather limited.

Therefore, this adds deprecation warnings to command line arguments for individual PEtab tables and for the `-y` `--yaml` flags, which are no longer required, and the removal of which will make `petablint` easier to use.

Related to https://github.com/PEtab-dev/libpetab-python/issues/271

### DIFF
--- a/petab/petablint.py
+++ b/petab/petablint.py
@@ -49,7 +49,7 @@ def parse_cli_args():
     )
 
     # Call with set of files
-    group = parser.add_argument_group("Check individual files *DEPRECATED*")
+    group = parser.add_argument_group("Check individual files *DEPRECATED*. Please contact us via GitHub issues, if you need this.")
     group.add_argument(
         "-s", "--sbml", dest="sbml_file_name", help="SBML model filename"
     )
--- a/petab/petablint.py
+++ b/petab/petablint.py
@@ -49,34 +49,35 @@ def parse_cli_args():
     )
 
     # Call with set of files
-    parser.add_argument(
+    group = parser.add_argument_group("Check individual files *DEPRECATED*")
+    group.add_argument(
         "-s", "--sbml", dest="sbml_file_name", help="SBML model filename"
     )
-    parser.add_argument(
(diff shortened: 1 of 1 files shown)


petab/petablint.py
```diff
@@ -87,13 +88,18 @@ def parse_cli_args():
...
         "-y",
         "--yaml",
+        dest="yaml_file_name_deprecated",
+        help="PEtab YAML problem filename. "
+        "*DEPRECATED* pass the file name as positional argument instead.",
+    )
+    group.add_argument(
         dest="yaml_file_name",
         help="PEtab YAML problem filename",
+        nargs="?",
```
dilpath, line 98: I wonder if it's possible to say `yaml_file_name_deprecated` takes 0 arguments, i.e. it's treated as just a "flag", and instead its argument then gets treated as a positional argument, and hence saved to `yaml_file_name`. Anyway, not important

dilpath (approved):
:+1:

petab/petablint.py
```diff
@@ -49,34 +49,35 @@ def parse_cli_args():
     )
 
     # Call with set of files
-    parser.add_argument(
+    group = parser.add_argument_group("Check individual files *DEPRECATED*")
```
dilpath, line 52: Could we add to this message, so that if someone happens to rely on individual files, we are informed?

```suggestion
    group = parser.add_argument_group("Check individual files *DEPRECATED*. Please contact us via GitHub issues, if you need this.")
```

petab/petablint.py
dweindl: It can be done a would slightly simplify the code, but the `petablint --help` output would be more confusing, I think.

PR #277, opened 2024-06-25
Doc: versioning and deprecation policy

dweindl:
:eyes: https://petab--277.org.readthedocs.build/projects/libpetab-python/en/277/development.html#


dilpath (approved):
:+1:

PR #278, opened 2024-06-26
Validator: check for positive bounds for log-scaled parameter

dweindl:
Check the parameter table for positive bounds for log-scaled estimated parameters that don't have an explicit intialization prior.

See discussion in #259

Supersedes and closes #259


petab/lint.py
```diff
@@ -557,6 +557,17 @@ def check_parameter_bounds(parameter_df: pd.DataFrame) -> None:
                     f"Bounds for {row[PARAMETER_SCALE]} scaled parameter "
                     f"{ row.name} must be positive."
                 )
+            if (
+                row.get(PARAMETER_SCALE, LIN) in [LOG, LOG10]
+                and (row[LOWER_BOUND] == 0.0 or row[UPPER_BOUND] == 0.0)
+                and not row.get(INITIALIZATION_PRIOR_TYPE)
+            ):
```
dilpath, line 564: I would move `row[UPPER_BOUND] == 0.0` above, i.e. change `row[UPPER_BOUND] < 0.0` to `row[UPPER_BOUND] <= 0.0` -- does not make sense to estimate something that can only take the value `0.0`.

dilpath (approved):
:+1:

petab/lint.py
dweindl: lb=ub is allowed so far. I'd say things that are legal but dumb shouldn't raise exceptions. Although we might want to add some optional hints for improvement at some point.

dilpath: Fine for me :+1: Should we check the initialization prior then, to make sure it's sampling in finite bounds?

PR #282, opened 2024-06-27
Introduce petab.v1 package

dweindl:
As discussed in #271, we need to accommodate the changes related to the upcoming PEtab v2. The goal is to move PEtab 1.0 functionality to a `petab.v1` subpackage and PEtab 2.0 functionality to a `petab.v2` subpackage.

This PR moves (almost) all code from the `petab` package to a `petab.v1` sub-package. Keeps all non-private objects importable from their previous location, but issues `DeprecationWarnings`.

**From the next release on, all consumers should change all `from petab[.$x] import $y` to `from petab.v1[.$x] import $y`.**

Fixes a couple of sphinx-issues that occurred on the way. Some obscure sphinx-failures remain, but they aren't critical.

Also adds some missing `__all__`s.


dilpath (approved):
Nice!

I looked at most files and didn't see any issues :+1: I looked at `petab/v1` less since I guess it's mostly copied.

Is it possible to include this in CI by running the tests with both like so?
```python
import petab as petabv0
import petab.v1

petab = petabv0
# run tests
petab = petab.v1
```

Should the tests also be split into `tests/v1` and `tests/v2`?

dweindl:
> I looked at `petab/v1` less since I guess it's mostly copied.

Yes.

> Is it possible to include this in CI by running the tests with both like so?
> 
> ```python
> import petab as petabv0
> import petab.v1
> 
> petab = petabv0
> # run tests
> petab = petab.v1
> ```

For now, I would keep running tests through the deprecated functions. If they pass there, I don't have much doubt that they will work through petab.v1, as effectively everything was moved there. Once we remove the old modules, I would switch the tests.

> Should the tests also be split into `tests/v1` and `tests/v2`?

Yes, I think that makes sense. Will do that separately, though.

PR #284, opened 2024-06-28
Separate v1 and v2 tests

dweindl:
Move tests to v1 and v2 subpackages as done for the library.

Also improve error handling for incorrect SBML models / paths.


dilpath (approved):
:+1:

PR #285, opened 2024-06-28
Add petab.v2.Problem

dweindl:
Copy of v1.Problem with all deprecated functionality removed.

Loading PEtab 2.0 problems with petab.v1.Problem is now deprecated.


dilpath (approved):
:+1:

dilpath:
Were `to_files` and `to_files_generic` deprecated / Is there a method to write a PEtab problem to disk, in `petab.v2`?

PR #287, opened 2024-06-28
Fix validation for remote files

### LINKED ISSUE #286: Validation fails for remote files
`validate_yaml_semantics` uses `os.path.isfile` to check whether referenced files are available and fails for remote files.

dweindl:
Actually load files instead of just checking `exists`, which failed on remote files.

Closes #286


dilpath (approved):
:+1:

PR #288, opened 2024-06-28
New validation API

dweindl:
For petab.v2, refactor the current validation setup to 
* be more modular
* allow PEtab extensions to register additional validation steps in the future
* make validation messages accessible programmatically (previously, they were logged directly)

Also:
* Make `petablint` work with both v1 and v2 problems.
* Add first validation rules that are specific to v2

### DIFF
--- a/petab/v2/lint.py
+++ b/petab/v2/lint.py
@@ -7,6 +7,9 @@ from dataclasses import dataclass
 from enum import IntEnum
 from pathlib import Path
 
+import numpy as np
+import pandas as pd
+
 from .problem import Problem
 
 logger = logging.getLogger(__name__)
@@ -45,57 +48,46 @@ class ValidationTask(ABC):
     """A task to validate a PEtab problem."""
 
     @abstractmethod
-    def run(self, problem: Problem) -> list[ValidationResult]:
+    def run(self, problem: Problem) -> ValidationResult | None:
         """Run the validation task.
 
         Arguments:
             problem: PEtab problem to check.
         Returns:
-            A list of validation results. Empty if no issues were found.
+            Validation results or ``None``
         """
         ...
+
+    def __call__(self, *args, **kwargs):
+        return self.run(*args, **kwargs)
 
 
 class ModelValidationTask(ValidationTask):
     """A task to validate the model of a PEtab problem."""
 
-    def run(self, problem: Problem) -> list[ValidationResult]:
+    def run(self, problem: Problem) -> ValidationResult | None:
         if problem.model is None:
-            return ValidationResultList(
-                [
-                    ValidationResult(
-                        level=ValidationEventLevel.WARNING,
-                        message="Model is missing.",
-                    )
-                ]
+            return ValidationResult(
+                level=ValidationEventLevel.WARNING,
+                message="Model is missing.",
             )
 
-        if problem.model.is_valid():
-            return []
-
-        # TODO get actual model validation messages
-        return ValidationResultList(
-            [
-                ValidationResult(
-                    level=ValidationEventLevel.ERROR,
-                    message="Model is invalid.",
-                )
-            ]
-        )
+        if not problem.model.is_valid():
+            # TODO get actual model validation messages
+            return ValidationResult(
+                level=ValidationEventLevel.ERROR,
+                message="Model is invalid.",
+            )
 
 
 class MeasurementTableValidationTask(ValidationTask):
     """A task to validate the measurement table of a PEtab problem."""
 
-    def run(self, problem: Problem) -> list[ValidationResult]:
+    def run(self, problem: Problem) -> ValidationResult | None:
         if problem.measurement_df is None or problem.measurement_df.empty:
-            return ValidationResultList(
-                [
-                    ValidationResult(
-                        level=ValidationEventLevel.ERROR,
-                        message="Measurement table is missing or empty.",
-                    )
-                ]
+            return ValidationResult(
+                level=ValidationEventLevel.ERROR,
+                message="Measurement table is missing or empty.",
             )
 
         try:
@@ -111,28 +103,19 @@ class MeasurementTableValidationTask(ValidationTask):
                     problem.measurement_df, problem.condition_df
                 )
         except AssertionError as e:
-            return ValidationResultList(
-                [
-                    ValidationResult(
-                        level=ValidationEventLevel.ERROR, message=str(e)
-                    )
-                ]
+            return ValidationResult(
+                level=ValidationEventLevel.ERROR, message=str(e)
             )
-        return []
 
 
 class ConditionTableValidationTask(ValidationTask):
     """A task to validate the condition table of a PEtab problem."""
 
-    def run(self, problem: Problem) -> list[ValidationResult]:
+    def run(self, problem: Problem) -> ValidationResult | None:
         if problem.condition_df is None or problem.condition_df.empty:
-            return ValidationResultList(
-                [
-                    ValidationResult(
-                        level=ValidationEventLevel.ERROR,
-                        message="Condition table is missing or empty.",
-                    )
-                ]
+            return ValidationResult(
+                level=ValidationEventLevel.ERROR,
+                message="Condition table is missing or empty.",
             )
 
         try:
@@ -145,14 +128,9 @@ class ConditionTableValidationTask(ValidationTask):
                 mapping_df=problem.mapping_df,
             )
         except AssertionError as e:
-            return ValidationResultList(
-                [
-                    ValidationResult(
-                        level=ValidationEventLevel.ERROR, message=str(e)
-                    )
-                ]
+            return ValidationResult(
+                level=ValidationEventLevel.ERROR, message=str(e)
             )
-        return []
 
 
 class ObservableTableValidationTask(ValidationTask):
@@ -160,13 +138,9 @@ class ObservableTableValidationTask(ValidationTask):
 
     def run(self, problem: Problem):
         if problem.observable_df is None or problem.observable_df.empty:
-            return ValidationResultList(
-                [
-                    ValidationResult(
-                        level=ValidationEventLevel.ERROR,
-                        message="Observable table is missing or empty.",
-                    )
-                ]
+            return ValidationResult(
+                level=ValidationEventLevel.ERROR,
+                message="Observable table is missing or empty.",
             )
 
         try:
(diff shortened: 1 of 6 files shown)


petab/v2/problem.py
```diff
@@ -69,6 +78,9 @@ def __init__(
         self.mapping_df: pd.DataFrame | None = mapping_df
         self.model: Model | None = model
         self.extensions_config = extensions_config or {}
+        self.validation_tasks: list[
+            ValidationTask
+        ] = default_validation_taks.copy()
```
dilpath, line 83:
```suggestion
        ] = default_validation_tasks.copy()
```

petab/v2/lint.py
```diff
@@ -0,0 +1,331 @@
...
+    INFO = 10
+    # WARNING: Warning message, potential issues
+    WARNING = 20
+    # ERROR: Error message, action required
+    ERROR = 30
+    # CRITICAL: Critical error message, stops further validation
+    CRITICAL = 40
+
+
+default_validation_taks = [
```
dilpath, line 324:
```suggestion
default_validation_tasks = [
```

petab/v2/lint.py
```diff
@@ -0,0 +1,549 @@
...
+
+class CheckMeasurementTable(ValidationTask):
+    """A task to validate the measurement table of a PEtab problem."""
+
+    def run(self, problem: Problem) -> ValidationIssue | None:
+        if problem.measurement_df is None:
+            return
+
+        try:
+            check_measurement_df(problem.measurement_df, problem.observable_df)
```
dilpath, line 216: Open issue to one day replace these with individual checks? No need if you think it's obvious / will happen naturally as we move away from depending on  `petab.v1` methods

petab/v2/lint.py
```diff
@@ -0,0 +1,549 @@
...
+
+            if problem.condition_df is not None:
+                # TODO: handle missing condition_df
```
dilpath, line 219: :+1: fine for now but I think some `DEBUG` or `INFO` message would be good, so a user knows not all aspects of the table were checked.

petab/v2/lint.py
```diff
@@ -0,0 +1,549 @@
...
+            for column_name in PARAMETER_DF_REQUIRED_COLS[
+                1:
+            ]:  # 0 is PARAMETER_ID
+                if not np.issubdtype(df[column_name].dtype, np.number):
+                    assert_no_leading_trailing_whitespace(
+                        df[column_name].values, column_name
+                    )
+
+            # nominal value is generally optional, but required if any for any
+            #  parameter estimate != 1
```
dilpath, line 308:
```suggestion
            # nominal value is required for non-estimated parameters
```

petab/v2/lint.py
```diff
@@ -0,0 +1,549 @@
...
+            non_estimated_par_ids = list(
+                df.index[
+                    (df[ESTIMATE] != 1)
+                    | (
+                        pd.api.types.is_string_dtype(df[ESTIMATE])
```
dilpath, line 313: :hankey:

Ideally this would be a validation error

petab/v2/lint.py
```diff
@@ -0,0 +1,549 @@
...
+                        f"but column {NOMINAL_VALUE} is missing."
+                    )
+                try:
+                    df.loc[non_estimated_par_ids, NOMINAL_VALUE].apply(float)
+                except ValueError:
+                    return ValidationError(
+                        f"Expected numeric values for `{NOMINAL_VALUE}` "
+                        "in parameter table "
+                        "for all non-estimated parameters."
+                    )
```
dilpath, line 333:
```suggestion
            # TODO implement as validators `assert_has_fixed_parameter_nominal_values`
            # and `assert_correct_table_dtypes`
            if non_estimated_par_ids:
                if NOMINAL_VALUE not in df:
                    return ValidationError(
                        "Parameter table contains parameters "
                        f"{non_estimated_par_ids} that are not "
                        "specified to be estimated, "
                        f"but column {NOMINAL_VALUE} is missing."
                    )
                try:
                    df.loc[non_estimated_par_ids, NOMINAL_VALUE].apply(float)
                except ValueError:
                    return ValidationError(
                        f"Expected numeric values for `{NOMINAL_VALUE}` "
                        "in parameter table "
                        "for all non-estimated parameters."
                    )
```

petab/v2/lint.py
```diff
@@ -0,0 +1,549 @@
...
+        # the mapping table
+        if missing and problem.mapping_df is not None:
+            model_to_petab_mapping = {}
+            for map_from, map_to in zip(
+                problem.mapping_df.index.values,
+                problem.mapping_df[MODEL_ENTITY_ID],
+                strict=True,
+            ):
+                if map_to in model_to_petab_mapping:
+                    model_to_petab_mapping[map_to].append(map_from)
```
dilpath, line 385: Why would multiple model IDs map to a single PEtab ID? Is it e.g. to observe the sum of all species regardless of isotope state in rule-based modeling? Then I would have more questions.. but irrelevant here.

petab/v2/lint.py
```diff
@@ -0,0 +1,549 @@
...
+        if problem.visualization_df is None:
+            return
+
+        from petab.visualize.lint import validate_visualization_df
+
+        if validate_visualization_df(problem):
+            return ValidationIssue(
+                level=ValidationIssueSeverity.ERROR,
+                message="Visualization table is invalid.",
+            )
```
dilpath, line 449: For this single-error cases, there could already be some `SingleValidationTask` class like we discussed, e.g.

```suggestion
class CheckVisualizationTable(SingleValidationTask):
    """A task to validate the visualization table of a PEtab problem."""
    level = ValidationIssueSeverity.ERROR
    message = "Visualization table is invalid."

    def run(self, problem: Problem) -> ValidationIssue | None:
        if problem.visualization_df is None:
            return

        from petab.visualize.lint import validate_visualization_df

        return (not?) validate_visualization_df(problem)
        or
        return self.handle(validation_visualization_df(problem))
```

petab/v2/lint.py
```diff
@@ -0,0 +1,549 @@
...
+
+
+def get_required_parameters_for_parameter_table(
```
dilpath, line 452: Didn't check, let me know if I should

dilpath (approved):
Looks great :+1:

petab/v2/lint.py
```diff
@@ -0,0 +1,549 @@
...
+from collections import OrderedDict
+from dataclasses import dataclass
+from enum import IntEnum
+from pathlib import Path
+
+import numpy as np
+import pandas as pd
+
+from petab.C import NOISE_PARAMETERS, OBSERVABLE_PARAMETERS
+from petab.v1 import (
```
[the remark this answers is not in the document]
dweindl: So far, my approach was reusing what is reusable, and once something changes, we add our own implementation in .v2. So, yes, I would keep the import from v1.

petab/v2/lint.py
```diff
@@ -0,0 +1,549 @@
...
+
+    def __str__(self):
+        return f"{self.level.name}: {self.message}"
+
+
+class ValidationError(ValidationIssue):
+    """A validation result with level ERROR."""
+
+    def __init__(self, message: str):
+        super().__init__(ValidationIssueSeverity.ERROR, message)
```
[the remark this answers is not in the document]
dweindl: But then we'd still have to override `__init__`, otherwise one can do `ValidationError(ValidationIssueSeverity.INFO, "bla")` which would be weird.

I think I'd rather stick to `__init__`, but validating `level` there would be good.

petab/v2/lint.py
```diff
@@ -0,0 +1,549 @@
...
+    """A task to check if a table exists in the PEtab problem."""
+
+    def __init__(self, table_name: str):
+        if table_name not in ["measurement", "observable", "parameter"]:
+            # all others are optional
+            raise ValueError(
+                f"Table name {table_name} is not supported. "
+                "Supported table names are 'measurement', 'observable', "
+                "'parameter'."
+            )
```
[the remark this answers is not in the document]
dweindl: I wouldn't use `ValidationIssueSeverity.*` here, because it's not really a issue with the to-be-validated petab problem, but rather an implementation issue inside the library here.

Actually, it's not caught in  `validate`, because `CheckTableExists` is already instantiated before. `validate` will only execute `.run()`.

petab/v2/lint.py
dweindl: This was about handling the (iirc) optional conditions table in petab2 (which means, all conditionIds will be empty and model values are to be used). This will require various changes. I guess I will open a separate issue on that.
So in this case it's not about something going unnoticed, but rather a false positive.

petab/v2/lint.py
dweindl: That's the old problem that data types are undefined. I guess we could change that in v2, but that's a separate discussion. We also have the issue, for example, in the conditions table. Depending on the contents, the columns will be object or float.

petab/v2/lint.py
dweindl: Iirc, the only reason is that so far, it's not explicitly forbidden.

petab/v2/lint.py
dweindl: The only change is the addition of ` and p not in problem.observable_df.index` in L513.

petab/v2/lint.py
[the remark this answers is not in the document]
dilpath: This would remove it from `__init__`
```python
@dataclass
class ValidationError(ValidationIssue):
    level: ValidationIssueSeverity = field(default=ValidationIssueSeverity.ERROR, init=False)
```

but also fine as is

petab/v2/problem.py
```diff
@@ -620,3 +632,35 @@ def n_priors(self) -> int:
             return 0
 
         return self.parameter_df[OBJECTIVE_PRIOR_PARAMETERS].notna().sum()
+
+    def validate(self) -> ValidationResultList:
```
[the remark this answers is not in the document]
dweindl: For now your suggestion would work equally well, but my plan is that at some point, when a problem is loaded, an extension plugin can add (or if really necessary, remove) checks on a specific Problem instance.

petab/v2/problem.py
```diff
@@ -620,3 +632,35 @@ def n_priors(self) -> int:
...
+        """Validate the PEtab problem."""
+        from ..v2.lint import ValidationIssueSeverity, ValidationResultList
```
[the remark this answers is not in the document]
dweindl: Circular import issues. Will double-check.

petab/v2/lint.py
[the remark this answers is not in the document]
dilpath: Oops, I missed that this was in `__init__`. Thanks!

petab/v2/lint.py
dilpath: Ah, then my comment is for the `if problem.*_df is None: return` in other validators.

dweindl: I'd say those are fine, because either the tables are required and `CheckTableExists` will emit an error, or those are optional. I am not sure if we want messages for missing optional tables, but I don't have a strong opinion there. If they aren't there but should be there, I'd say this will most likely cause downstream errors.
For validation of v1 problems we have a warning on missing visualization tables, but I regret having added that :sweat_smile:.

petab/v2/lint.py
```diff
@@ -0,0 +1,550 @@
...
+            problem.model,
+            mapping_df=problem.mapping_df,
+            **formula_type,
+        )
+        placeholders = get_placeholders(
+            problem.observable_df,
+            **placeholder_sources,
+        )
+        for p in output_parameters:
+            if p not in placeholders and p not in problem.observable_df.index:
```
dilpath, line 514:
```suggestion
            # remove `p not in problem.observable_df.index` when
            # `get_output_parameters` is updated for PEtab v2/v1.1, where
            # observable IDs are allowed in observable formulae
            if p not in placeholders and p not in problem.observable_df.index:
```

If `parameter_ids` was a set, we could do `parameter_ids |= output_parameters - placeholders`. We could `return sorted(parameter_ids)` to simplify the code a bit. The `dict` makes the code a little weird but fine as is.

petab/v2/lint.py
dilpath: Could be better to implement this like so then
```python
class CheckConditionTable(ValidationTask):
    requires=[CheckTableExists("condition")]
    ...
```

and remove `CheckTableExists("condition")` from `default_validation_tasks`. This would implement the dependency graph though -- fine to keep out of this PR

petab/v2/problem.py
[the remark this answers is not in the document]
dweindl: done

petab/versions.py
```diff
@@ -0,0 +1,56 @@
...
+        yaml_config = load_yaml(problem)
+        version = yaml_config.get(FORMAT_VERSION)
+    elif isinstance(problem, dict):
+        version = problem.get(FORMAT_VERSION)
+    else:
+        raise ValueError(f"Unsupported argument type: {type(problem)}")
+
+    version = str(version)
+    if version.startswith("2.") or version == "2":
+        return True
```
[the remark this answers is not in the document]
dweindl: I think `get_major_version` alone would be sufficient. :+1:

petab/v2/lint.py
dweindl: https://github.com/PEtab-dev/libpetab-python/issues/289

petab/v2/lint.py
dweindl: https://github.com/PEtab-dev/libpetab-python/issues/290

petab/v2/lint.py
dweindl: I don't recall the reason for insisting on the given order, but I don't see a problem with requiring `sorted` for deterministic results. Also currently the type annotation is wrong. Will change.

petab/v2/lint.py
dweindl: Yes, a dependency graph would be convenient to run as many checks as possible in a single run. However, I think maintaining that properly will be tedious. For now, I'd leave it as is.

petab/v2/lint.py
dweindl: I am not sure this will be more intuitive. People will always wonder whether `true` indicates passed or failed. Where the message is context-specific, this wouldn't really save anything. Will leave as is for now.

dilpath: Ah, I intended `self.handle` to return `ValidationIssue` or `None` like done elsewhere in this PR

dilpath: > Where the message is context-specific

Are you saying, an error message for `SingleValidationTasks` might be context-specific? Doesn't that contradict the idea of a `SingleValidationTask`?

dweindl: I meant that instead of `There are log-scaled parameters with negative bounds.`, we want to say `There are log-scaled parameters with negative bounds: parameter1, parameter42.`, so one would have to modify self.message again somewhere, and wouldn't really gain much.

dilpath: I see, but not if `self.handle` also accepts an optional level/message. Anyway, not important

PR #292, opened 2024-06-29
Doc: fix formatting / missing type annotations

dweindl:
:eyes: https://petab--292.org.readthedocs.build/projects/libpetab-python/en/292/modules.html

### DIFF
--- a/petab/models/sbml_model.py
+++ b/petab/models/sbml_model.py
@@ -1,4 +1,4 @@
-"""Deprecated module. Use petab.pysb_model instead."""
+"""Deprecated module. Use petab.v1.models.sbml_model instead."""
 from petab import _deprecated_import_v1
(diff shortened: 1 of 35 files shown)


petab/models/sbml_model.py
```diff
@@ -1,3 +1,4 @@
+"""Deprecated module. Use petab.pysb_model instead."""
```
dilpath, line 1: # :raised_eyebrow:

```suggestion
"""Deprecated module. Use petab.v1.models.sbml_model instead."""
```

dilpath (approved):
:+1:

petab/models/sbml_model.py
dweindl: :scream_cat:

PR #309, opened 2024-09-24
Function for replacing `objectivePrior*` by observables/measurements

### LINKED ISSUE #307: Function for replacing `objectivePrior*` by observables/measurements
Some frameworks might not support parameter priors out of the box. In such cases, it would be convenient if one could just automatically modify the PEtab problem and replace the `objectivePrior*` entries by extra observables and measurements that will result in an equivalent objective function.

dweindl:
Reformulate the given PEtab problem such that the objective priors are converted to measurements. This is done by adding a new observable for each prior and adding a corresponding measurement to the measurement table. The resulting optimization problem will be equivalent to the original problem. This is meant to be used for tools that do not support priors.

Closes #307

### DIFF
--- a/petab/v1/priors.py
+++ b/petab/v1/priors.py
@@ -42,6 +42,14 @@ def priors_to_measurements(problem: Problem):
     The new measurement is the prior distribution itself. The resulting
     optimization problem will be equivalent to the original problem.
     This is meant to be used for tools that do not support priors.
+    
+    The conversion involves the probability density function (PDF) of the prior, the parameters (e.g. location and scale) of that prior PDF, and the scale and value of the estimated parameter. Currently, `uniform` priors are not supported by this method. This method creates observables with:
+    - `observableFormula`: the parameter value on the `parameterScale`
+    - `observableTransformation`: `log` for `logNormal`/`logLaplace` distributions, `lin` otherwise
+
+    and measurements with:
+    - `measurement`: the PDF location
+    - `noiseFormula`: the PDF scale
 
     Arguments
     ---------
@@ -70,7 +78,7 @@ def priors_to_measurements(problem: Problem):
         if parameter_scale == LIN:
             return parameter_id
         if parameter_scale == LOG:
-            return f"log({parameter_id})"
+            return f"ln({parameter_id})"
         if parameter_scale == LOG10:
             return f"log10({parameter_id})"
         raise ValueError(f"Unknown parameter scale {parameter_scale}.")
--- a/petab/v1/priors.py
+++ b/petab/v1/priors.py
@@ -4,6 +4,7 @@ import copy
 import numpy as np
 import pandas as pd
 
+from ..v2.C import PREEQUILIBRATION_CONDITION_ID
 from . import (
     ESTIMATE,
     LAPLACE,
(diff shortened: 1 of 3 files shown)


petab/v1/priors.py
```diff
@@ -0,0 +1,173 @@
...
+        ] = np.nan
+
+    if new_observable_dicts:
+        new_problem.observable_df = pd.concat(
+            [
+                pd.DataFrame(new_observable_dicts).set_index(OBSERVABLE_ID),
+                new_problem.observable_df,
+            ]
+        )
+    if new_measurement_dicts:
```
dilpath, line 165: The top of this method checks that there is some prior, so these `if` statements will always be `True`?

tests/v1/test_priors.py
```diff
@@ -0,0 +1,116 @@
...
+    for parameter_id, prior in zip(parameter_ids, priors, strict=True):
+        prior_type, prior_pars, par_scale, par_bounds = prior
+        if prior_type == petab.v1.PARAMETER_SCALE_NORMAL:
+            prior_contrib += norm.logpdf(
+                petab_problem_priors.x_nominal_free_scaled[
+                    petab_problem_priors.x_free_ids.index(parameter_id)
+                ],
+                loc=prior_pars[0],
+                scale=prior_pars[1],
+            )
```
dilpath, line 104: Looks useful, could be moved to some `calculate_posterior_for_table` method like `calculate_llh_for_table`

tests/v1/test_priors.py
```diff
@@ -0,0 +1,116 @@
...
+            #  evaluating the prior contribution. until then, two test
+            #  problems should suffice
+            assert problem_id == "Raimundez_PCB2020"
+            pytest.skip(f"Prior type {prior_type} not implemented")
+
+    assert np.isclose(
+        llh_priors + prior_contrib, llh_measurements, rtol=1e-3, atol=1e-16
+    ), (llh_priors + prior_contrib, llh_measurements)
+    # check that the tolerance is not too high
+    assert np.abs(prior_contrib) > 1e-3 * np.abs(llh_priors)
```
dilpath, line 116: Can also ensure that the original problem was not changed by `priors_to_measurements` (or, that the new and old problems differ in some way)

tests/v1/test_priors.py
```diff
@@ -0,0 +1,116 @@
...
+    assert (
+        petab_problem_measurements.observable_df.shape[0]
+        > petab_problem_priors.observable_df.shape[0]
+    )
+    assert (
+        petab_problem_measurements.measurement_df.shape[0]
+        > petab_problem_priors.measurement_df.shape[0]
+    )
+
+    # verity that the objective function value is the same
```
dilpath, line 46:
```suggestion
    # verify that the objective function value is the same
```

petab/v1/priors.py
```diff
@@ -0,0 +1,173 @@
...
+def priors_to_measurements(problem: Problem):
+    """Convert priors to measurements.
+
+    Reformulate the given problem such that the objective priors are converted
+    to measurements. This is done by adding a new observable
+    ``prior_{parameter_id}`` for each estimated parameter that has an objective
+    prior, and adding a corresponding measurement to the measurement table.
+    The new measurement is the prior distribution itself. The resulting
+    optimization problem will be equivalent to the original problem.
+    This is meant to be used for tools that do not support priors.
```
dilpath, line 44: Just to explain the logic, since this is unusual with possibly multiple approaches.

```suggestion
    This is meant to be used for tools that do not support priors.
    
    The conversion involves the probability density function (PDF) of the prior, the parameters (e.g. location and scale) of that prior PDF, and the scale and value of the estimated parameter. Currently, `uniform` priors are not supported by this method. This method creates observables with:
    - `observableFormula`: the parameter value on the `parameterScale`
    - `observableTransformation`: `log` for `logNormal`/`logLaplace` distributions, `lin` otherwise

    and measurements with:
    - `measurement`: the PDF location
    - `noiseFormula`: the PDF scale
```

dilpath (approved):
Overall, looks great, thanks!

petab/v1/priors.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # add measurement
+        # we can just use any condition and time point since the parameter
+        # value is constant
+        sim_cond_id = new_problem.condition_df.index[0]
+
+        new_measurement_dicts.append(
+            {
+                OBSERVABLE_ID: new_obs_id,
+                TIME: 0,
```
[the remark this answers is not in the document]
dweindl: Yes, that makes sense. That will also avoid any extra solver stops in tools that don't check for time-dependency of the outputs.

tests/v1/test_priors.py
dweindl: I fully agree. However, I would like to have some proper `Prior` class instead of those current tuples before extending the priors API. I would rather keep that for another PR. I will open an issue.

PR #317, opened 2024-11-06
Preserve column ordering in `priors_to_measurements`

dweindl:
Appending instead of preprending the new observables/measurements will preserve the column order.


dilpath (approved):
:+1:

PR #318, opened 2024-11-07
Make model id optional for PySBModel

dweindl:
If no ID is given, use `pysb.Model.name` as fallback. As is already the case for SbmlModel and its `id` attribute.


dilpath (approved):
:+1:

PR #326, opened 2024-12-03
Store problem configuration in Problem

### LINKED ISSUE #324: Path management
When generating a `petab.Problem` from a yaml file, all files are automatically prefixed by the location of the yaml file. However when directly passing the corresponding `dict` (potentially after manipulation), it is no longer possible to specify a prefix and the only way for the import to work correctly is to `os.chdir` to the appropriate directory.

Straightforward to add a corresponding command line argument, but I guess it would be generally great to store a prefix path somewhere in the generated `petab.Problem`, as this would make it easier to reference files within the problem (not part of the current spec afaik, but it would be great to have in the future).

Opinions?

dweindl:
Introduces `Problem.config` which contains the info from the PEtab yaml file.

Sometimes it is convenient to have the original filenames around.

Pydantic gives more helpful error messages than `jsonschema` in case of incorrect inputs. Later on, this could replace  `jsonschema` completely.

If accepted, I will add the same for `petab.v2.Problem`.

Related to #324.

### DIFF
--- a/petab/v1/problem.py
+++ b/petab/v1/problem.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING
 from warnings import warn
 
 import pandas as pd
+from pydantic import AnyUrl, BaseModel, Field, RootModel
 
 from . import (
     conditions,
@@ -79,6 +80,7 @@ class Problem:
         observable_df: pd.DataFrame = None,
         mapping_df: pd.DataFrame = None,
         extensions_config: dict = None,
+        config: ProblemConfig = None,
     ):
         self.condition_df: pd.DataFrame | None = condition_df
         self.measurement_df: pd.DataFrame | None = measurement_df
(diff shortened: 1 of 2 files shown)


dilpath (approved):
Looks good, could also add filename metadata to the specific table objects, so that e.g. the top-half of the measurements is saved to `measurements1.tsv` and the bottom-half to `measurements2.tsv`. Fine to merge already

dweindl:
> could also add filename metadata to the specific table objects, so that e.g. the top-half of the measurements is saved to `measurements1.tsv` and the bottom-half to `measurements2.tsv`. Fine to merge already


Yeah, I was thinking along the same lines. The problem is, we'd probably have to store the filenames in a new column, which people might not like. Storing it elsewhere, will cause issues in case the dataframe gets modified.

PR #327, opened 2024-12-03
Enable passing the base path to Problem.from_yaml

### LINKED ISSUE #324: Path management

dweindl:
When passing the problem configuration as `dict` to `Problem.from_yaml`, one should be able to specify the base path for resolving relative paths. See #324.

Closes #324

### DIFF
--- a/petab/v1/problem.py
+++ b/petab/v1/problem.py
@@ -251,21 +251,28 @@ class Problem:
         )
 
     @staticmethod
-    def from_yaml(yaml_config: dict | Path | str) -> Problem:
+    def from_yaml(
+        yaml_config: dict | Path | str, base_path: str | Path = None
+    ) -> Problem:
         """
         Factory method to load model and tables as specified by YAML file.
 
         Arguments:
             yaml_config: PEtab configuration as dictionary or YAML file name
+            base_path: Base directory or URL to resolve relative paths
         """
         if isinstance(yaml_config, Path):
             yaml_config = str(yaml_config)
 
-        get_path = lambda filename: filename  # noqa: E731
         if isinstance(yaml_config, str):
-            path_prefix = get_path_prefix(yaml_config)
+            if base_path is None:
+                base_path = get_path_prefix(yaml_config)
             yaml_config = yaml.load_yaml(yaml_config)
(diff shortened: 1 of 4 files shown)


petab/v1/problem.py
```diff
@@ -251,21 +251,28 @@ def from_files(
...
-            path_prefix = get_path_prefix(yaml_config)
+            if base_path is None:
+                base_path = get_path_prefix(yaml_config)
             yaml_config = yaml.load_yaml(yaml_config)
-            get_path = lambda filename: f"{path_prefix}/{filename}"  # noqa: E731
+
+        def get_path(filename):
+            if base_path is None:
+                return filename
+            return f"{base_path}/{filename}"
```
dilpath, line 275: Is this safe across platforms? Alternative
```suggestion
            return Path(base_path) / filename
```

petab/v2/problem.py
```diff
@@ -117,24 +117,31 @@ def __str__(self):
...
+                base_path = get_path_prefix(yaml_file)
+            yaml_config = yaml.load_yaml(yaml_file)
         else:
             yaml_file = None
-            get_path = lambda filename: filename  # noqa: E731
+
+        def get_path(filename):
+            if base_path is None:
+                return filename
+            return f"{base_path}/{filename}"
```
dilpath, line 144: Also here
```suggestion
            return Path(base_path) / filename
```

dilpath (approved):
:+1:

petab/v1/problem.py
dweindl: Yes, it's tested on Windows on GHA. `Path` won't work for remote files on Windows afaik. PosixPath might work.

dilpath: Alright, fine for me then :+1:

Issue #220, opened 2023-07-25
API for creating petab problems from scratch


dweindl:
It would be nice to have some high-level API to create PEtab problems from scratch, instead of manually assembling the dataframes.

Something like
```python
pp = petab.Problem()
pp.add_condition("conditionId", param1=2, param2=3)
pp.add_observable("observable1", formula="x1", ...)
pp.add_measurement(observable_id="observable1", time=2.5, measurement=1.337, noise_parameters=0.1, ...)
pp.add_parameter(parameter_id="p1", estimated=False, nominal_value=42)
```

dweindl:
Done

PR #328, opened 2024-12-04
Functions for adding conditions/observables/parameter to Problem

dweindl:
Add functions for adding individual conditions/observables/parameter/measurements to Problem. This will simplify writing test cases and interactively assembling petab problems.

`petab.v2.Problem.add_*` will be added / updated to the new format separately.

Related to #220.

### DIFF
--- a/petab/v1/problem.py
+++ b/petab/v1/problem.py
@@ -3,8 +3,9 @@ from __future__ import annotations
 
 import os
 import tempfile
-from collections.abc import Iterable
+from collections.abc import Iterable, Sequence
 from math import nan
+from numbers import Number
 from pathlib import Path, PurePosixPath
 from typing import TYPE_CHECKING
 from warnings import warn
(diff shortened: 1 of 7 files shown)


petab/v1/problem.py
```diff
@@ -1005,3 +1006,177 @@ def n_priors(self) -> int:
             return 0
 
         return self.parameter_df[OBJECTIVE_PRIOR_PARAMETERS].notna().sum()
+
+    def add_condition(self, id_: str, name: str = None, **kwargs):
```
dilpath, line 1010: To match column headers/`C.py`? Not exactly the same due to underscores... but we could decide whether to go for consistent or context-specific IDs everywhere.
```suggestion
    def add_condition(self, condition_id: str, condition_name: str = None, **kwargs):
```

petab/v1/problem.py
```diff
@@ -1005,3 +1006,177 @@ def n_priors(self) -> int:
...
+            self.condition_df = pd.concat([self.condition_df, tmp_df])
+
+    def add_observable(
+        self,
+        id_: str,
+        formula: str | float | int,
+        noise_formula: str | float | int = None,
+        noise_distribution: str = None,
+        transform: str = None,
+        name: str = None,
```
dilpath, line 1034: As above
```suggestion
        observable_id: str,
        observable_formula: str | float | int,
        noise_formula: str | float | int = None,
        noise_distribution: str = None,
        transform: str = None,
        observable_name: str = None,
```

petab/v1/problem.py
```diff
@@ -1005,3 +1006,177 @@ def n_priors(self) -> int:
...
+            record[NOISE_DISTRIBUTION] = [noise_distribution]
+        if transform is not None:
+            record[OBSERVABLE_TRANSFORMATION] = [transform]
+        record.update(kwargs)
+
+        tmp_df = pd.DataFrame(record).set_index([OBSERVABLE_ID])
+        if self.observable_df is None:
+            self.observable_df = tmp_df
+        else:
+            self.observable_df = pd.concat([self.observable_df, tmp_df])
```
dilpath, line 1067: `pd.concat` docs:
> Any None objects will be dropped silently unless they are all None in which case a ValueError will be raised.
```suggestion
        tmp_df = pd.DataFrame(record).set_index([OBSERVABLE_ID])
        self.observable_df = pd.concat([self.observable_df, tmp_df])
```

petab/v1/problem.py
dilpath: Alternative: change all of these `add_*` methods to take `**kwargs` that are used to create a `pd.Series`, which is then validated and concatenated. `get_*_df` can be used to set the index. Then no table-specific code, and no need to redefine column names here?
- `if not isinstance(kwarg, str) and isinstance(kwarg, list): kwarg = PARAMETER_SEPARATOR.join(map(str, kwarg))`

petab/v2/petab1to2.py
```diff
@@ -136,6 +136,9 @@ def _copy_file(src: Path | str, dest: Path | str):
     src = str(src)
     dest = str(dest)
 
+    if src == dest:
+        return
```
dilpath, line 140: Is this to skip computation or avoid an error? If the latter, then `src`/`dest` could be relative, so `Path.resolve()` should be used first.

dilpath (commented):
I think the new code here might already be handled by the old code, so I left a comment to discuss that suggestion.

petab/v1/problem.py
dweindl: Yeah, I was struggling with what would be preferable. In `add_condition`, it feels redundant to prefix everything with `condition_`. Then again, it might be considered confusing if the arguments don't match the table columns. For me, the former felt more important. Also with regards to potentially introducing a proper object model, I think we'd want things to be more pythonic and less petaby.

I don't think the kwargs-solution would be very convenient. That would mean, you'd have write `add_observable(**{petab.OBSERVABLE_ID:"foo", petab.SIMULATION_CONDITION_ID: "bar"})`, I think then I rather directly go back to constructing dataframes.

petab/v1/problem.py
dweindl: There is still some unnecessary things happening in pandas before this not-None object is returned. Then again, these functions here aren't so much about efficiency. I will think about it, thanks for the suggestion.

petab/v2/petab1to2.py
dweindl: It's the latter. Agreed, resolve would be preferable. I didn't use it, because src/dest could be URLs. But on second thought, `shutil.copy` would break on those anyways...

dweindl: done

petab/v1/problem.py
dweindl: I'll leave it as is for. As discussed elsewhere, the v2 API is likely to change drastically overall where these points will be addressed.

dilpath: Alright, fine for v1, we can revisit v2 in case we change the column names there

dilpath (approved):

PR #331, opened 2024-12-05
Add `SbmlModel.from_antimony`

dweindl:
Simplify creating a PEtab SbmlModel from antimony files / strings.

Replace simplesbml by antimony in tests.

### DIFF
--- a/tests/v1/test_parameter_mapping.py
+++ b/tests/v1/test_parameter_mapping.py
@@ -32,16 +32,15 @@ class TestGetSimulationToOptimizationParameterMapping:
             }
         )
 
-        import simplesbml
+        model = SbmlModel.from_antimony(
+            "dynamicParameter1 = 1.0; "
+            "dynamicParameter2 = 2.0; "
+            "dynamicParameter3 = 3.0; "
+            # add species, which will have initial concentration in condition
+            #  table but which should not show up in mapping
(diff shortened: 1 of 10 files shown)


tests/v1/test_parameter_mapping.py
```diff
@@ -364,15 +358,14 @@ def test_partial_override(condition_df_2_conditions):
...
-        ss_model.addParameter("observableParameter2_obs1", 0.0)
-        ss_model.addParameter("observableParameter1_obs2", 0.0)
+        model = SbmlModel.from_antimony(
+            "fixedParameter1 = 0.5\n"
+            "fixedParameter2 = 1.0\n"
+            "dynamicParameter1 = 0.0\n"
+            "observableParameter1_obs1 = 0.0\n"
+            "observableParameter2_obs1 = 0.0\n"
+            "observableParameter1_obs2 = 0.0"
+        )
```
dilpath, line 368: Use `"""` instead of `\n`? Or does antimony interpret indentation?

dilpath (approved):
Nice, add support for antimony to the docs and spec in v2?

tests/v1/test_parameter_mapping.py
dweindl: Makes sense. No, it doesn't.

PR #332, opened 2024-12-05
Support for petab v2 experiments

dweindl:
Add basic support for PEtab version 2 experiments (see also https://github.com/PEtab-dev/PEtab/issues/586, and  https://github.com/PEtab-dev/PEtab/pull/581). Follow-up to #334.

Partially supersedes #263, which was started before petab.v1/petab.v2 were introduced and before https://github.com/PEtab-dev/PEtab/issues/586.

* updates the required fields in the measurement table
* updates some validation functions to not expect the old `simulationConditionId`s (but does not do full validation yet)
* extends PEtab v1 up-conversion to create a new experiment table.

### DIFF
--- a/petab/v2/petab1to2.py
+++ b/petab/v2/petab1to2.py
@@ -131,7 +131,7 @@ def petab1to2(yaml_config: Path | str, output_dir: Path | str = None):
                     {
                         v2.C.EXPERIMENT_ID: exp_id,
                         v2.C.CONDITION_ID: preeq_cond_id,
-                        v2.C.TIME: float("-inf"),
+                        v2.C.TIME: v2.C.TIME_PREEQUILIBRATION,
                     }
                 )
             experiments.append(
--- a/petab/v2/petab1to2.py
+++ b/petab/v2/petab1to2.py
@@ -4,8 +4,8 @@ from contextlib import suppress
 from itertools import chain
 from pathlib import Path
 from urllib.parse import urlparse
+from uuid import uuid4
 
-import numpy as np
 import pandas as pd
 from pandas.io.common import get_handle, is_url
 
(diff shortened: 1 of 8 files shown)


petab/v2/petab1to2.py
```diff
@@ -98,10 +98,81 @@ def petab1to2(yaml_config: Path | str, output_dir: Path | str = None):
...
 
+        # records for the experiment table to be created
+        experiments = []
+
+        def create_experiment_id(sim_cond_id: str, preeq_cond_id: str) -> str:
+            if not sim_cond_id and not preeq_cond_id:
+                return ""
+            if preeq_cond_id:
+                preeq_cond_id = f"{preeq_cond_id}_"
+            exp_id = f"experiment_{preeq_cond_id}{sim_cond_id}"
```
dilpath, line 109: Might be nicer since single `_` is common in condition names (e.g. many look like `model1_data1` in the benchmark collection).
```suggestion
            exp_id = f"experiment__{preeq_cond_id}__{sim_cond_id}"
```

petab/v2/petab1to2.py
```diff
@@ -98,10 +98,81 @@ def petab1to2(yaml_config: Path | str, output_dir: Path | str = None):
...
+            if exp_id in experiments:  # noqa: B023
+                i = 1
+                while f"{exp_id}_{i}" in experiments:  # noqa: B023
+                    i += 1
+                exp_id = f"{exp_id}_{i}"
```
dilpath, line 114: Why would this occur? Ah, because no symbol is inserted between `preeq_cond_id` and `sim_cond_id`? Then maybe my suggestion with `__` means you can remove this?

petab/v2/petab1to2.py
```diff
@@ -98,10 +98,81 @@ def petab1to2(yaml_config: Path | str, output_dir: Path | str = None):
...
+            #  combination
+            sim_cond_id = row[v1.C.SIMULATION_CONDITION_ID]
+            preeq_cond_id = row.get(v1.C.PREEQUILIBRATION_CONDITION_ID, "")
+            exp_id = create_experiment_id(sim_cond_id, preeq_cond_id)
+            if preeq_cond_id:
+                experiments.append(
+                    {
+                        v2.C.EXPERIMENT_ID: exp_id,
+                        v2.C.CONDITION_ID: preeq_cond_id,
+                        v2.C.TIME: float("-inf"),
```
dilpath, line 134: Could use `-TIME_STEADY_STATE`, or add some `TIME_PREEQUILIBRATION`

dilpath (approved):
Looks good!

petab/v2/petab1to2.py
dweindl: There might be measurements for `('foo', 'bar')` and `(None, 'foo__bar')` which would result in colliding experiment IDs. You never know...

petab/v2/petab1to2.py
dweindl: Added `TIME_PREEQUILIBRATION`

PR #333, opened 2024-12-05
Sbml model enhancements

fbergmann:
This PR allows to create SBML models from string, but more critically considers an SBML model with warnings still as valid SbmlModel. The current PEtab implementation considers a file with warnings as invalid SBML, which is not the case.


dweindl (approved):
Thanks, looks good to me.

PR #336, opened 2024-12-08
Fix test_priors_to_measurements

dweindl:
Fixes an issue in `test_priors_to_measurements` which led to evaluating the prior at the wrong parameter values (using the location parameter of the prior instead of the actually estimated parameters). The problem was in the test code, not the tested code.

### DIFF
--- a/tests/v1/test_priors.py
+++ b/tests/v1/test_priors.py
@@ -8,7 +8,15 @@ import pytest
 from scipy.stats import norm
 
 import petab.v1
-from petab.v1 import get_simulation_conditions
+from petab.v1 import (
+    ESTIMATE,
+    MEASUREMENT,
+    OBJECTIVE_PRIOR_TYPE,
+    OBSERVABLE_ID,
+    SIMULATION,
+    get_simulation_conditions,
+    get_simulation_df,
+)
 from petab.v1.priors import priors_to_measurements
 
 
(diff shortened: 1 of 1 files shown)


tests/v1/test_priors.py
```diff
@@ -17,20 +25,29 @@
...
         assert petab.v1.lint_problem(petab_problem_priors) is False
+
     original_problem = deepcopy(petab_problem_priors)
+    x_scaled_dict = dict(
+        zip(
+            original_problem.x_free_ids,
+            original_problem.x_nominal_free_scaled,
+            strict=True,
+        )
+    )
```
dilpath, line 48: Should it really be scaled values here? I assume the prior is either (1) unscaled or (2) on `parameterScale`. The former means this should be unscaled, and the latter is already handled in the observable formulae for these dummy prior observables, so no scaling is required here?

dweindl: So far, this test only handles PARAMETER_SCALE_NORMAL (to be extended in #329). Yes, the scaling is handled in the observable formulae, but here, we need to write the simulation table. For that, we do not evaluate the observable formula using the given parameters, but write the result directly. Therefore, we need the scaled ones.

tests/v1/test_priors.py
```diff
@@ -17,20 +25,29 @@
...
-
     if problem_id == "Isensee_JCB2018":
         # required to match the stored simulation results below
         petab.v1.flatten_timepoint_specific_output_overrides(
             petab_problem_priors
         )
         assert petab.v1.lint_problem(petab_problem_priors) is False
+
     original_problem = deepcopy(petab_problem_priors)
+    x_scaled_dict = dict(
```
dilpath, line 42:
```suggestion
    # All priors in this test case are defined on parameter scale, hence
    # the dummy measurements will take the scaled nominal values.
    x_scaled_dict = dict(
```

PR #338, opened 2024-12-09
Store problem configuration in `v2.Problem`

dweindl:
Introduces `v2.Problem.config` which contains the info from the PEtab yaml file. The same as #326, but for `v2.Problem`.

### DIFF
--- a/petab/v2/problem.py
+++ b/petab/v2/problem.py
@@ -12,6 +12,7 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 import pandas as pd
+from pydantic import AnyUrl, BaseModel, Field
 
 from ..v1 import (
     conditions,
@@ -25,6 +26,7 @@ from ..v1 import (
     yaml,
 )
 from ..v1.models.model import Model, model_factory
+from ..v1.problem import ListOfFiles, VersionNumber
 from ..v1.yaml import get_path_prefix
 from ..v2.C import *  # noqa: F403
 from . import experiments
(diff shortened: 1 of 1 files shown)


petab/v2/problem.py
```diff
@@ -199,40 +202,38 @@ def get_path(filename):
...
-        if len(problem0[MODEL_FILES] or []) > 1:
+        if len(problem0.model_files or []) > 1:
             # TODO https://github.com/PEtab-dev/libpetab-python/issues/6
             raise NotImplementedError(
                 "Support for multiple models is not yet implemented."
             )
-        if not problem0[MODEL_FILES]:
+        if not problem0.model_files:
             model = None
         else:
```
dilpath, line 228:
```suggestion
        model = None
        if problem0.model_files:
```

petab/v2/problem.py
```diff
@@ -981,3 +976,49 @@ def add_experiment(self, id_: str, *args):
...
+
+class ProblemConfig(BaseModel):
+    """The PEtab problem configuration."""
+
+    filepath: str | AnyUrl | None = Field(
+        None,
+        description="The path to the PEtab problem configuration.",
+        exclude=True,
+    )
+    base_path: str | AnyUrl | None = Field(
```
dilpath, line 1016: I prefer just `root` but fine as is

dilpath (approved):
:+1: Fine to merge as-is

petab/v2/problem.py
dweindl: My association with `root` would be that the other files are somewhere below in the hierarchy, and `root/..` would be illegal. This is not the case here. It might be just my association, though. I am open to change it later on if others will strong about it.

dilpath: `root` is used as the location to which all other things are relative here:
https://docs.python.org/3/library/shutil.html#shutil.make_archive

But I think others will have varying opinions about the difference between `root` and `base` so fine :+1:

PR #340, opened 2024-12-11
Fix `get_required_parameters_for_parameter_table`

dweindl:
The previous check did not make any sense. See discussion at https://github.com/PEtab-dev/libpetab-python/pull/339#discussion_r1878856556.

Also fix a missing import.

### DIFF
--- a/petab/v1/parameters.py
+++ b/petab/v1/parameters.py
@@ -332,9 +332,8 @@ def get_required_parameters_for_parameter_table(
         if not model.has_entity_with_id(p):
             parameter_ids[p] = None
 
-    # remove parameters that occur in the condition table and are overridden
-    #  for ALL conditions
-    for p in condition_df.columns[~condition_df.isnull().any()]:
+    # parameters that are overridden via the condition table are not allowed
+    for p in condition_df.columns:
         try:
             del parameter_ids[p]
         except KeyError:
--- a/petab/v2/lint.py
+++ b/petab/v2/lint.py
@@ -621,11 +621,8 @@ def get_required_parameters_for_parameter_table(
         if not problem.model.has_entity_with_id(p)
     )
 
-    # remove parameters that occur in the condition table and are overridden
-    #  for ALL conditions
-    for p in problem.condition_df.columns[
-        ~problem.condition_df.isnull().any()
-    ]:
+    # parameters that are overridden via the condition table are not allowed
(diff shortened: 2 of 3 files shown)


petab/v1/parameters.py
```diff
@@ -332,9 +332,8 @@ def append_overrides(overrides):
         if not model.has_entity_with_id(p):
             parameter_ids[p] = None
 
-    # remove parameters that occur in the condition table and are overridden
-    #  for ALL conditions
-    for p in condition_df.columns[~condition_df.isnull().any()]:
+    # parameters that are overridden via the condition table are not allowed
+    for p in condition_df.columns:
```
dilpath, line 336: Instead of try-except, could do one of these, but fine as is
```python
for p in set(condition_df.columns).intersection(parameter_ids):
    del parameter_ids[p]

for p in condition_df.columns:
    if p in parameter_ids:
        del parameter_ids[p]
```

petab/v2/lint.py
```diff
@@ -621,11 +621,8 @@ def append_overrides(overrides):
         if not problem.model.has_entity_with_id(p)
     )
 
-    # remove parameters that occur in the condition table and are overridden
-    #  for ALL conditions
-    for p in problem.condition_df.columns[
-        ~problem.condition_df.isnull().any()
-    ]:
+    # parameters that are overridden via the condition table are not allowed
+    for p in problem.condition_df.columns:
```
dilpath, line 625: Since parameter IDs is now a set, can change try-except/loop to
```python
parameter_ids -= set(condition_df.columns)
```

dilpath (approved):
Thanks!

petab/v1/parameters.py
dweindl: Right, but then one has to look up each element twice.

petab/v2/lint.py
dweindl: Since it's right before returning, it could be done just on the keys, but then future modifications might easily mess up the ordering. I'd leave it as is until Python has a proper OrderedSet.

dilpath: Where does it become `OrderedDict`? In the v1 code it's initialized as an `OrderedDict`, but here as a `set`?

dweindl: You are right. I was in v1.

PR #341, opened 2024-12-12
Fix Prior.from_par_dict for missing priorParameters columns

dweindl:
Previously, missing `*PriorParameters` would have resulted in a KeyError.


m-philipps (approved):
Thanks!

Issue #337, opened 2024-12-09
Use proper object model for petab.v2


dweindl:
For petab version 2, we will have to rewrite a large part of the current functionality, which gives us the opportunity to rethink the API. For v2, @dilpath and I would like to change the DataFrame based API to a proper object model (e.g., instead of `observable_df: pd.DataFrame`, we'd have `observables: list[Observable]`).
This will greatly simplify validation and dealing with default values as well setting proper data types.

fbergmann:
we could still have an observable_df property that serialises the observables to df's for those that need it, might also be handy for serialisation later

PR #345, opened 2024-12-18
v2: Observable/conditions/experiments/... objects

dweindl:
Implement an object model for the various PEtab entities as an alternative to working with the plain DataFrames.

Related to #337.

Object names are very likely to change. This is more of a proof of concept.

Also adapt to recent changes in https://github.com/PEtab-dev/PEtab/pull/581

### DIFF
--- a/petab/v2/C.py
+++ b/petab/v2/C.py
@@ -130,30 +130,23 @@ CONDITION_ID = "conditionId"
 CONDITION_NAME = "conditionName"
 #: Column in the condition table with the ID of an entity that is changed
 TARGET_ID = "targetId"
-#: Column in the condition table with the type of value that is changed
-VALUE_TYPE = "valueType"
+#: Column in the condition table with the operation type
+OPERATION_TYPE = "operationType"
 #: Column in the condition table with the new value of the target entity
 TARGET_VALUE = "targetValue"
-# value types:
-VT_CONSTANT = "constant"
-VT_INITIAL = "initial"
-VT_RATE = "rate"
-VT_ASSIGNMENT = "assignment"
-VT_RELATIVE_RATE = "relativeRate"
-VT_RELATIVE_ASSIGNMENT = "relativeAssignment"
-VALUE_TYPES = [
-    VT_CONSTANT,
-    VT_INITIAL,
-    VT_RATE,
-    VT_ASSIGNMENT,
-    VT_RELATIVE_RATE,
-    VT_RELATIVE_ASSIGNMENT,
+# opeartion types:
+OT_CUR_VAL = "setCurrentValue"
+OT_NO_CHANGE = "noChange"
+
+OPERATION_TYPES = [
+    OT_CUR_VAL,
+    OT_NO_CHANGE,
 ]
 
 CONDITION_DF_COLS = [
     CONDITION_ID,
     TARGET_ID,
-    VALUE_TYPE,
+    OPERATION_TYPE,
     TARGET_VALUE,
 ]
 
--- a/petab/v2/C.py
+++ b/petab/v2/C.py
@@ -134,7 +134,7 @@ TARGET_ID = "targetId"
 OPERATION_TYPE = "operationType"
 #: Column in the condition table with the new value of the target entity
 TARGET_VALUE = "targetValue"
-# opeartion types:
+# operation types:
 OT_CUR_VAL = "setCurrentValue"
 OT_NO_CHANGE = "noChange"
 
--- a/petab/v2/core.py
+++ b/petab/v2/core.py
@@ -279,14 +279,27 @@ class ExperimentsTable(BaseModel):
     experiments: list[Experiment]
 
     @classmethod
-    def from_dataframe(cls, df: pd.DataFrame) -> ExperimentsTable:
+    def from_dataframe(
+        cls, df: pd.DataFrame, conditions: ConditionsTable = None
+    ) -> ExperimentsTable:
         if df is None:
             return cls(experiments=[])
 
-        experiments = [
-            Experiment(**row.to_dict())
-            for _, row in df.reset_index().iterrows()
-        ]
+        if conditions is None:
+            conditions = {}
+
+        experiments = []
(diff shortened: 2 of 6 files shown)


petab/v2/C.py
```diff
@@ -129,30 +129,23 @@
...
-VT_RELATIVE_RATE = "relativeRate"
-VT_RELATIVE_ASSIGNMENT = "relativeAssignment"
-VALUE_TYPES = [
-    VT_CONSTANT,
-    VT_INITIAL,
-    VT_RATE,
-    VT_ASSIGNMENT,
-    VT_RELATIVE_RATE,
-    VT_RELATIVE_ASSIGNMENT,
+# opeartion types:
```
dilpath, line 136:
```suggestion
# operation types:
```

petab/v2/C.py
```diff
@@ -129,30 +129,23 @@
...
+OT_CUR_VAL = "setCurrentValue"
+OT_NO_CHANGE = "noChange"
```
dilpath, line 138: Other constants in C.py are not shortened, so could stick to that "convention". To reduce the length, `OT`/`OPERATION_TYPE` could be dropped from the start of the name, since I don't think we have that style elsewhere either.

petab/v2/core.py
```diff
@@ -0,0 +1,711 @@
...
+    """
+
+    NORMAL = C.NORMAL
+    LAPLACE = C.LAPLACE
+    UNIFORM = C.UNIFORM
+    LOG_NORMAL = C.LOG_NORMAL
+    LOG_LAPLACE = C.LOG_LAPLACE
+    PARAMETER_SCALE_NORMAL = C.PARAMETER_SCALE_NORMAL
+    PARAMETER_SCALE_LAPLACE = C.PARAMETER_SCALE_LAPLACE
+    PARAMETER_SCALE_UNIFORM = C.PARAMETER_SCALE_UNIFORM
```
dilpath, line 111: Looks like this belongs in v2/C.py, as possibly also remove `C.PRIOR_TYPES` (or deprecate and define it in terms of `ObjectivePriorType`. Is there a reason to have both `ObjectivePriorType` and `InitializationPriorType`?

petab/v2/core.py
```diff
@@ -0,0 +1,711 @@
...
+        "name",
+        "formula",
+        "noise_formula",
+        "noise_formula",
+        "noise_distribution",
+        "transformation",
+        mode="before",
+    )
+    @classmethod
+    def convert_nan_to_none(cls, v, info: ValidationInfo):
```
dilpath, line 147:
```suggestion
    def convert_nan_to_default(cls, v, info: ValidationInfo):
```

petab/v2/core.py
```diff
@@ -0,0 +1,711 @@
...
+            return cls(observables=[])
+
+        observables = [
+            Observable(**row.to_dict())
+            for _, row in df.reset_index().iterrows()
+        ]
+
+        return cls(observables=observables)
+
+    def to_dataframe(self) -> pd.DataFrame:
```
dilpath, line 191: Fine as is, but since df is common enough
```suggestion
    def from_df(cls, df: pd.DataFrame) -> ObservablesTable:
        if df is None:
            return cls(observables=[])

        observables = [
            Observable(**row.to_dict())
            for _, row in df.reset_index().iterrows()
        ]

        return cls(observables=observables)

    def to_df(self) -> pd.DataFrame:
```

petab/v2/core.py
```diff
@@ -0,0 +1,711 @@
...
+            return None
+
+        return sympify_petab(v)
+
+    class Config:
+        populate_by_name = True
+        arbitrary_types_allowed = True
+
+
+class ObservablesTable(BaseModel):
```
dilpath, line 167: I guess this PR is independent of any decision on choice between "observable table" and "observables table" naming style, or is this the decision?

petab/v2/core.py
```diff
@@ -0,0 +1,711 @@
...
+        return self
+
+
+class OperationType(str, Enum):
+    """Operation types for model changes in the PEtab conditions table."""
+
+    # TODO update names
+    SET_CURRENT_VALUE = "setCurrentValue"
+    NO_CHANGE = "noChange"
+    ...
```
dilpath, line 223: Same with this, can replace the "same" thing in C.py

petab/v2/core.py
dilpath, line 1: There is a lot of repeated code (e.g. `from/to_dataframe/tsv`, and `add/iadd`). Is it nicer/possible to somehow only define these once?

dilpath (approved):
Looks good! I reviewed this without considering the other PEtab v2 things, i.e. I only checked that the OOP code look good, since I guess those other things are out-of-scope here.

petab/v2/C.py
dweindl: I would prefer introducing prefixes elsewhere than dropping them. I find that this makes it clearer where they belong.

petab/v2/core.py
dweindl: I don't know if they will always be the same. Not strong opinion.

petab/v2/core.py
dweindl: Independent. Happy to rename just about anything here after the v2 specs are finalized.

petab/v2/C.py
dweindl: Also fine to completely replace them by enums.

petab/v2/core.py
dweindl: I expect that some additional validation will happen there in the future which would make it rather difficult to implement that only once. I think the classes will me more understandable if they have their own add/iadd instead of some complex function combining the logic of all the different types.

petab/v2/core.py
dilpath: Alright, while they are the same then I think this makes sense
```python
class PriorType:
    ...

ObjectivePriorType = PriorType
InitializationPriorType = PriorType
```

petab/v2/C.py
dweindl: I'll leave it for now, as there is a good chance that "operation types" will be dropped completely

PR #346, opened 2024-12-18
Fix v2 import Deprecation warning

dweindl:
Fixes this nonsensical warning:

> E   DeprecationWarning: Accessing `petab.v2` is deprecated and will be removed in the next major release. Please use `petab.v1.v2` instead.


dilpath (approved):
petab.v2 it is then...

PR #347, opened 2024-12-18
petab.calculate: compare all common columns

dweindl:
For computing residuals, ... from measurement + simulation tables, we need to match the corresponding rows. Previously, this was done using a subset of PEtab measurement table columns and checking whether all values in these columns match.

This changes it to using the full set of overlapping columns, not only the known measurement columns.
With that, the same functions can be used for PEtab v2 measurement/simulation tables.


dilpath (approved):
:+1: 

Could make sense to only check official PEtab columns and avoid custom ones, but I don't see a reason why the custom columns would differ either, so fine :+1:

PR #348, opened 2024-12-18
Problem.add_parameter: Rename kwarg estimated->estimate

dweindl:
Change the `estimated` argument in Problem.add_parameter to `estimate` to match the column name.

### DIFF
--- a/petab/v1/problem.py
+++ b/petab/v1/problem.py
@@ -1105,9 +1105,7 @@ class Problem:
             PARAMETER_ID: [id_],
         }
         if estimate is not None:
-            record[ESTIMATE] = [
-                int(estimate) if isinstance(estimate, bool | int) else estimate
-            ]
+            record[ESTIMATE] = [int(estimate)]
         if nominal_value is not None:
             record[NOMINAL_VALUE] = [nominal_value]
         if scale is not None:
--- a/petab/v1/problem.py
+++ b/petab/v1/problem.py
@@ -1075,7 +1075,7 @@ class Problem:
         self,
         id_: str,
         estimate: bool | str | int = True,
-        nominal_value=None,
+        nominal_value: Number | None = None,
         scale: str = None,
         lb: Number = None,
         ub: Number = None,
--- a/petab/v1/problem.py
+++ b/petab/v1/problem.py
@@ -1074,7 +1074,7 @@ class Problem:
     def add_parameter(
         self,
         id_: str,
-        estimated: bool | str | int = True,
+        estimate: bool | str | int = True,
         nominal_value=None,
         scale: str = None,
         lb: Number = None,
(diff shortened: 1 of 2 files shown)


petab/v1/problem.py
```diff
@@ -1074,7 +1074,7 @@ def add_observable(
     def add_parameter(
         self,
         id_: str,
-        estimated: bool | str | int = True,
+        estimate: bool | str | int = True,
         nominal_value=None,
```
dilpath, line 1078:
```suggestion
        nominal_value: Number | None = None,
```

petab/v2/problem.py
```diff
@@ -829,7 +829,7 @@ def add_observable(
     def add_parameter(
         self,
         id_: str,
-        estimated: bool | str | int = True,
+        estimate: bool | str | int = True,
         nominal_value=None,
```
dilpath, line 833:
```suggestion
        nominal_value: Number | None = None,
```

petab/v2/problem.py
```diff
@@ -859,11 +859,9 @@ def add_parameter(
         record = {
             PARAMETER_ID: [id_],
         }
-        if estimated is not None:
+        if estimate is not None:
             record[ESTIMATE] = [
-                int(estimated)
-                if isinstance(estimated, bool | int)
-                else estimated
+                int(estimate) if isinstance(estimate, bool | int) else estimate
```
dilpath, line 864:
```suggestion
                int(estimate) if isinstance(estimate, bool | str) else estimate
```

petab/v1/problem.py
```diff
@@ -1104,11 +1104,9 @@ def add_parameter(
         record = {
             PARAMETER_ID: [id_],
         }
-        if estimated is not None:
+        if estimate is not None:
             record[ESTIMATE] = [
-                int(estimated)
-                if isinstance(estimated, bool | int)
-                else estimated
+                int(estimate) if isinstance(estimate, bool | int) else estimate
```
dilpath, line 1109:
```suggestion
                int(estimate) if isinstance(estimate, bool | str) else estimate
```

dilpath: I think probably fine to replace this with simply `[int(estimate)]`

dilpath (approved):
:+1:

petab/v1/problem.py
dweindl: Right. None is handled above.

dilpath: Hm, None is not in the signature

dweindl: Ah, yeah, because it's required in the parameters table.
That still an open question how much validation should happen in `Problem.add_*(...)`.

dilpath: I think none (in v2 at least), since the upcoming pydantic object will validate upon construction

PR #350, opened 2024-12-20
Fix version checks

### LINKED ISSUE #349: v2: version check
`pteab.v2.Problem.from_yaml` will only accept "2.0.0" as version string, even though it sounds like "2" would also be valid according to spec.

see https://github.com/PEtab-dev/libpetab-python/blob/99373c4340512def5a23bf2a75805c3f88251251/petab/v2/problem.py#L139

dweindl:
Add function for parsing version numbers. Update checks.

Fixes #349.


dilpath (approved):
:+1:

PR #353, opened 2024-12-20
Update/use ProblemConfig

dweindl:
* add to_yaml
* refactor model, avoid extra elements in object tree
* use ProblemConfig in petab1to2

### DIFF
--- a/petab/v1/problem.py
+++ b/petab/v1/problem.py
@@ -11,7 +11,7 @@ from typing import TYPE_CHECKING
 from warnings import warn
 
 import pandas as pd
-from pydantic import AnyUrl, BaseModel, Field, RootModel
+from pydantic import AnyUrl, BaseModel, Field
 
 from ..versions import get_major_version
 from . import (
@@ -1185,33 +1185,14 @@ class Problem:
         )
 
 
-class VersionNumber(RootModel):
-    root: str | int
-
-
-class ListOfFiles(RootModel):
-    """List of files."""
-
-    root: list[str | AnyUrl] = Field(..., description="List of files.")
-
-    def __iter__(self):
-        return iter(self.root)
-
-    def __len__(self):
-        return len(self.root)
-
-    def __getitem__(self, index):
-        return self.root[index]
-
-
 class SubProblem(BaseModel):
     """A `problems` object in the PEtab problem configuration."""
 
-    sbml_files: ListOfFiles = []
-    measurement_files: ListOfFiles = []
-    condition_files: ListOfFiles = []
-    observable_files: ListOfFiles = []
-    visualization_files: ListOfFiles = []
+    sbml_files: list[str | AnyUrl] = []
+    measurement_files: list[str | AnyUrl] = []
+    condition_files: list[str | AnyUrl] = []
+    observable_files: list[str | AnyUrl] = []
+    visualization_files: list[str | AnyUrl] = []
 
 
 class ProblemConfig(BaseModel):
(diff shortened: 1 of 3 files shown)


petab/v1/problem.py
```diff
@@ -1185,33 +1185,14 @@ def add_measurement(
...
-
 class SubProblem(BaseModel):
     """A `problems` object in the PEtab problem configuration."""
 
-    sbml_files: ListOfFiles = []
-    measurement_files: ListOfFiles = []
-    condition_files: ListOfFiles = []
-    observable_files: ListOfFiles = []
-    visualization_files: ListOfFiles = []
+    sbml_files: list[str | AnyUrl] = []
```
dilpath, line 1191: This can be `list[Path | AnyUrl]`, since pydantic has native support for `pathlib.Path` IIRC. But then you might need to make a custom path object that serializes to string https://github.com/dilpath/mkstd/blob/fb4ebfda629b0fb9eedb1def5ab05438ab85696e/mkstd/types/path.py

Fine as is, just noting in case it becomes useful later

petab/v1/problem.py
```diff
@@ -1227,6 +1208,16 @@ class ProblemConfig(BaseModel):
...
-    format_version: VersionNumber = 1
+    format_version: str | int = 1
     parameter_file: str | AnyUrl | None = None
     problems: list[SubProblem] = []
+
+    def to_yaml(self, filename: str | Path):
+        """Write the configuration to a YAML file.
+
+        filename: Destination file name. The parent directory will be created
+            if necessary.
```
dilpath, line 1219: Not in `__all__` so I guess it doesn't matter for docs, but usually this style in this repo
```suggestion
        Parameters:
            filename:
                Destination file name. The parent directory will be
                created if necessary.
```
.. I think. In the files touched by this PR alone, I see
```
Parameters
----------

Arguments:

Parameters:
```

petab/v2/problem.py
```diff
@@ -1024,7 +1023,17 @@ class ProblemConfig(BaseModel):
         description="The base path to resolve relative paths.",
         exclude=True,
     )
-    format_version: VersionNumber = "2.0.0"
+    format_version: str = "2.0.0"
```
dilpath, line 1026: `str | int` in v1 -- intentional difference?

petab/v2/problem.py
```diff
@@ -1024,7 +1023,17 @@ class ProblemConfig(BaseModel):
...
     parameter_file: str | AnyUrl | None = None
     problems: list[SubProblem] = []
     extensions: list[ExtensionConfig] = []
+
+    def to_yaml(self, filename: str | Path):
+        """Write the configuration to a YAML file.
+
+        filename: Destination file name. The parent directory will be created
```
dilpath, line 1034: Same doc comment re: `Parameters:`

dilpath (approved):
:+1:

petab/v1/problem.py
dweindl: Yeah, it would make sense to start a style guide.... 

I'd prefer `:param ...: `

petab/v2/problem.py
dweindl: That's how it is in the current schema draft. Although not strictly enforced so far, I think it would make sense to always require major.minor. But that's probably a separate discussion.

dilpath: Since there is no benefit to keeping `1` as an int, could use `str` instead of `str | int`, to use pydantic to handle the type casting.

Requiring a full version string is fine for me, but then we need to provide schemas for each minor version. Makes sense to support minor updates to the format, so I'm also in favor of expecting a full `X.X` or `X.X.X` string. major.minor` as suggested is probably enough since we probably don't need to distinguish "minor" and "patch" updates

petab/v1/problem.py
dilpath: Agreed, but no preference for the specific style

petab/v2/problem.py
dweindl: Agreed, but I'll keep that for a future update.

petab/v1/problem.py
dweindl: Started a style guide in https://github.com/PEtab-dev/libpetab-python/pull/354

petab/v2/problem.py
dweindl: Added the `format_version_tuple` property for a normalized version number number.

PR #358, opened 2025-03-12
doc: intersphinx, tox env

dweindl:
* Simplify linking to the PEtab specs
* Add tox env for building the documentation

### DIFF
--- a/petab/v1/problem.py
+++ b/petab/v1/problem.py
@@ -42,7 +42,9 @@ __all__ = ["Problem"]
 
 class Problem:
     """
-    PEtab parameter estimation problem as defined by
+    PEtab parameter estimation problem.
+
+    A PEtab problem as defined by:
 
     - model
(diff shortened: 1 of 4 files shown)


dweindl:
RTD builds pass. Remaining test failures are fixed in #345

petab/v1/problem.py
```diff
@@ -50,7 +52,9 @@ class Problem:
     - observables table
     - mapping table
```
dilpath, line 53: No mapping in v1? Fine for this PR

dilpath (approved):

petab/v1/problem.py
dweindl: No mapping in v1 in principle, but it was added there long ago for initial v2 support before we decided to have a separate `v2.Problem`. We should keep it there until petab.v2 has converged.

PR #360, opened 2025-03-13
doc: petab.v2.core

dweindl:
:eyes: https://petab--360.org.readthedocs.build/projects/libpetab-python/en/360/build/_autosummary/petab.v2.core.html

### DIFF
--- a/petab/v2/core.py
+++ b/petab/v2/core.py
@@ -134,7 +134,7 @@ class Observable(BaseModel):
 
     @field_validator("id")
     @classmethod
-    def validate_id(cls, v):
+    def _validate_id(cls, v):
         if not v:
             raise ValueError("ID must not be empty.")
         if not is_valid_identifier(v):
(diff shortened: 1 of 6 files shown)


petab/v2/core.py
```diff
@@ -219,12 +242,24 @@ class Change(BaseModel):
 
     A change to the model or model state, corresponding to an individual
     row of the PEtab conditions table.
+
+    >>> Change(
+    ...     target_id="k1",
+    ...     operation_type=OperationType.SET_CURRENT_VALUE,
+    ...     target_value="10",
```
dilpath, line 249: Using `"10"` is fine but it looks like int/float is handled? Is this to stress that the type is an expression? Then maybe using e.g. `k1+3` makes the point better.

dilpath (approved):
Looks good :+1: 

The "module attributes" (`ObjectivePriorType`, `InitializationPriorType`) are listed on RTD but not doc'd -- is this because they might be refactored? Fine for me.

petab/v2/core.py
dweindl: That's to test that strings are accepted/converted.

dweindl:
> The "module attributes" (`ObjectivePriorType`, `InitializationPriorType`) are listed on RTD but not doc'd -- is this because they might be refactored? Fine for me.

That's because they are just aliases. Either we have separate types, or we only have the members of the original type documented. And yes, I wouldn't bother until it's decided whether/how they will change.

PR #361, opened 2025-03-17
Update petab.v2

dweindl:
Adapt petab.v2 to the updated specs draft.

### DIFF
--- a/petab/v2/core.py
+++ b/petab/v2/core.py
@@ -411,17 +411,21 @@ class ExperimentPeriod(BaseModel):
 
     #: The start time of the period in time units as defined in the model.
     start: float = Field(alias=C.TIME)
+    # TODO: decide if optional
     #: The ID of the condition to be applied at the start time.
     condition_id: str = Field(alias=C.CONDITION_ID)
 
     #: :meta private:
     model_config = ConfigDict(populate_by_name=True)
 
-    @field_validator("condition_id")
+    @field_validator("condition_id", mode="before")
     @classmethod
     def _validate_id(cls, condition_id):
-        if not condition_id:
-            raise ValueError("ID must not be empty.")
+        # TODO to be decided if optional
+        if pd.isna(condition_id):
+            return ""
+        # if not condition_id:
+        #     raise ValueError("ID must not be empty.")
         if not is_valid_identifier(condition_id):
             raise ValueError(f"Invalid ID: {condition_id}")
         return condition_id
--- a/petab/v2/lint.py
+++ b/petab/v2/lint.py
@@ -545,6 +545,8 @@ class CheckExperimentConditionsExist(ValidationTask):
         missing_conditions = set(required_conditions) - set(
             existing_conditions
         )
+        # TODO NA allowed?
+        missing_conditions = {x for x in missing_conditions if not pd.isna(x)}
         if missing_conditions:
             return ValidationError(
                 f"Experiment table contains conditions that are not present "
@@ -842,7 +844,8 @@ default_validation_tasks = [
     CheckExperimentTable(),
     CheckValidPetabIdColumn("measurement", EXPERIMENT_ID, ignore_nan=True),
     CheckValidPetabIdColumn("experiment", EXPERIMENT_ID),
-    CheckValidPetabIdColumn("experiment", CONDITION_ID),
+    # TODO: NAN allowed?
+    CheckValidPetabIdColumn("experiment", CONDITION_ID, ignore_nan=True),
     CheckExperimentConditionsExist(),
     CheckObservableTable(),
     CheckObservablesDoNotShadowModelEntities(),
--- a/petab/v2/problem.py
+++ b/petab/v2/problem.py
@@ -798,10 +798,12 @@ class Problem:
             {
                 CONDITION_ID: id_,
                 TARGET_ID: target_id,
-                OPERATION_TYPE: op_type,
-                TARGET_VALUE: target_value,
+                OPERATION_TYPE: "setCurrentValue",
+                TARGET_VALUE: target_value
+                if not isinstance(target_value, tuple)
+                else target_value[1],
             }
-            for target_id, (op_type, target_value) in kwargs.items()
+            for target_id, target_value in kwargs.items()
         ]
         # TODO: is the condition name supported in v2?
         if name is not None:


matthiaskoenig:
@dweindl Is this usable already? I want to implement the timecourses in my tools but would need the updated petab library for that. Is this the right branch and working. If not what is still missing here?

dweindl:
> @dweindl Is this usable already? I want to implement the timecourses in my tools but would need the updated petab library for that. Is this the right branch and working. If not what is still missing here?

I'd suggest to wait another week. The whole `petab.v2` should be considered experimental. There's a lot in motion.

I am currently aligning things with https://github.com/PEtab-dev/PEtab/pull/581 and switching from DataFrame based functions to pydantic-objects-based ones for `petab.v2`.

matthiaskoenig:
Perfect.

petab/v2/core.py
```diff
@@ -43,6 +49,43 @@
 ]
 
 
+def _is_finite_or_neg_inf(v: float, info: ValidationInfo) -> float:
```
dilpath, line 52: Move these methods to the lint module? I guess they might be used there too, or is all validation intended to take place in the constructor of these pydantic objects?

petab/v2/core.py
```diff
@@ -200,85 +269,82 @@ def from_df(cls, df: pd.DataFrame) -> ObservablesTable:
...
+                else str(obs)
+                if not obs.is_number
+                else float(obs)
+            )
+            record[C.NOISE_FORMULA] = (
+                None
+                if noise is None
+                else str(noise)
+                if not noise.is_number
+                else float(noise)
```
dilpath, line 291: Some repeated code, could be moved to some `serialize_formula` method

petab/v2/core.py
```diff
@@ -745,13 +884,60 @@ def _validate_id(cls, v):
...
+            if v == 0:
+                return False
+            if v == 1:
+                return True
+
+        if isinstance(v, str):
+            if v == "0":
+                return False
+            if v == "1":
+                return True
```
dilpath, line 903: Sufficient? Could also add `True` and `False` to remove the above code too...
```suggestion
        if v in [0, 1, "0", "1"]:
            return bool(int(v))
```

petab/v2/lint.py
```diff
@@ -613,23 +551,116 @@ class CheckValidParameterInConditionOrParameterTable(ValidationTask):
...
+        }
+
+        disallowed_in_condition = {
+            x
+            for x in (entities_in_condition_table - allowed_in_condition_cols)
+            # we only check model entities here, not output parameters
+            if problem.model.has_entity_with_id(x)
+        }
+        if disallowed_in_condition:
+            is_or_are = "is" if len(disallowed_in_condition) == 1 else "are"
```
dilpath, line 592: :pinched_fingers:

tests/v2/test_conversion.py
```diff
@@ -36,10 +36,10 @@ def test_benchmark_collection(problem_id):
     logging.basicConfig(level=logging.DEBUG)
 
     if problem_id == "Froehlich_CellSystems2018":
+        # this is mostly about 6M sympifications in the condition table
         pytest.skip("Too slow. Re-enable once we are faster.")
```
dilpath, line 40: Not sure but maybe caching helps, e.g. https://github.com/AMICI-dev/AMICI/pull/1672/files

dilpath (approved):
Looks good so far!

petab/v1/math/sympify.py
```diff
@@ -26,14 +31,22 @@ def sympify_petab(expr: str | int | float) -> sp.Expr | sp.Basic:
         The sympy expression corresponding to `expr`.
         Boolean values are converted to numeric values.
     """
+    if isinstance(expr, sp.Expr):
+        # TODO: check if only PEtab-compatible symbols and functions are used
+        return expr
```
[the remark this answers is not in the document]
dweindl: You are right. It is not necessarily true, and it is part of the TODO.
After #362, I would just resympify the expression here, until there is a need for something more performant.

petab/v2/core.py
```diff
@@ -43,6 +49,43 @@
 ]
 
 
+def _is_finite_or_neg_inf(v: float, info: ValidationInfo) -> float:
+    if not np.isfinite(v) and v != -np.inf:
```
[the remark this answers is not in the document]
dweindl: Are you aware of inf-representations that would not work with the current form, but would work with `np.isneginf`? If not, I'd leave it to save the extra argument type checking that is probably happening in the latter.

petab/v2/core.py
dweindl: That is still somewhat of an open question. I wouldn't want to duplicate the validation logic. However, the question is, which validations might have to be made optional at some point to work with future extensions. Probably too worry about that though.

So far, I would see everything that concerns the files and the presence of columns, as well as everything that cannot be validated inside a specific pydantic model to be in the lint module.

Another open question related to validation is whether it should still be possible to construct the pydantic objects from an invalid table.

petab/v2/core.py
dweindl: I think that should work. But meanwhile https://github.com/PEtab-dev/PEtab/pull/614 happened. Needs updating.

Extra whitespace is not allowed, except for math expressions, right?

tests/v2/test_conversion.py
dweindl: I had the feeling that caching *is* the problem here (inside sympy), but that needs further investigation.

petab/v2/core.py
dilpath: No strong opinion but I guess that could be useful for e.g. loading partial problems into Paul's PEtab GUI

petab/v2/core.py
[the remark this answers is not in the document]
dweindl: This won't catch NaN

petab/v2/core.py
dilpath: > Extra whitespace is not allowed, except for math expressions, right?

Sounds fine to me. I don't see an issue either way, so whichever is more convenient

petab/v2/core.py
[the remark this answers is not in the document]
dilpath: Ah true, then good :+1:

petab/v2/core.py
dweindl: Same here. To be discussed at https://github.com/PEtab-dev/PEtab/.

PR #364, opened 2025-03-20
Add petab-compatible sympy string-printer

### LINKED ISSUE #362: Add petab-compatible sympy codeprinter
We need some codeprinter to stringify sympy expressions in a petab-compatible way. 
For example, we need to avoid `str(sympy.sympify("x^2"))` -> `'x**2'`.

dweindl:
Add a sympy Printer to stringify sympy expressions in a petab-compatible way. For example, we need to avoid `str(sympy.sympify("x^2"))` -> `'x**2'`.

Closes #362.

### DIFF
--- a/tests/v1/math/test_math.py
+++ b/tests/v1/math/test_math.py
@@ -6,9 +6,9 @@ import pytest
 import sympy as sp
 import yaml
 from sympy.abc import _clash
-from sympy.logic.boolalg import Boolean
+from sympy.logic.boolalg import Boolean, BooleanFalse, BooleanTrue
 
-from petab.math import sympify_petab
+from petab.v1.math import petab_math_str, sympify_petab
 
 
 def test_sympify_numpy():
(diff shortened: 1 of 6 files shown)


tests/v1/math/test_math.py
```diff
@@ -24,6 +24,21 @@ def test_parse_simple():
     assert float(sympify_petab("1 + 2 * (3 + 4) / 2")) == 8
 
 
+def test_assumptions():
+    # in PEtab, all symbols are expected to be real-valued
+    assert sympify_petab("x").is_real
+
+    # non-real symbols are changed to real
+    # TODO: should we raise an error instead?
+    assert sympify_petab(sp.Symbol("x", real=False)).is_real
```
dilpath, line 33: Probably raise an error, or does it occur in standard use?

dilpath (approved):
:+1:

tests/v1/math/test_math.py
dweindl: It happens if you pass `sympify("x")`. Not sure what standard use will be :sweat_smile:

dilpath: Ah, fine for me on the level of symbols. I was thinking of expressions, which I guess is irrelevant here, so nevermind.

PR #365, opened 2025-03-20
Add `evaluate: bool` argument to math parser

### LINKED ISSUE #363: Add `evaluate: bool` argument to math parser
Currently, when math expressions are parsed, they are evaluated as far as possible. This is not always desirable. Make this optional as far as possible.

dweindl:
So far, when math expressions were parsed, they were evaluated as far as possible. This was not always desirable.

Now, this optional as far as conveniently possible.

Closes #363.


dilpath (approved):
Looks good :+1:

PR #371, opened 2025-03-31
 Add SbmlModel.{to_antimony,to_sbml_str}

dweindl:
Simplify conversion of SBML models to antimony and XML strings. Convenient in particular for debugging.

### DIFF
--- a/petab/v1/models/sbml_model.py
+++ b/petab/v1/models/sbml_model.py
@@ -65,9 +65,7 @@ class SbmlModel(Model):
 
         # libsbml stuff cannot be serialized directly
         if self.sbml_model:
-            sbml_document = self.sbml_model.getSBMLDocument()
-            sbml_writer = libsbml.SBMLWriter()
-            state["sbml_string"] = sbml_writer.writeSBMLToString(sbml_document)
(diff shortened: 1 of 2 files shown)


dilpath (approved):
Looks fine to me. Should `Model` have `to_str` and `from_str` methods such that `Problem.model_dump` can serialize the model, too?

matthiaskoenig:
Can the functions be `from_antimony` and `to_antimony`. We have other converters to convert to SBML formats which are also string based. I would be more explicit here.

dilpath:
> Can the functions be `from_antimony` and `to_antimony`. We have other converters to convert to SBML formats which are also string based. I would be more explicit here.

Agreed, and this is the current state in this PR too. My comment was rather about (de)serialization of SBML and antimony models, not interconversion.

Since the `Model` base class already has `from_file` and `to_file`, I guess to would suffice to have some `from_str` and `to_str` with the file contents as a string.

But this (de)serialization to string is not important to me -- just that it would make the pydantic `model_dump` in this other PR more comprehensive.
https://github.com/PEtab-dev/libpetab-python/pull/372

PR #372, opened 2025-04-03
Add v2.Problem.model_dump

dweindl:
Convert a Problem to a dict.

### DIFF
--- a/petab/v2/problem.py
+++ b/petab/v2/problem.py
@@ -10,7 +10,7 @@ from collections.abc import Sequence
 from math import nan
 from numbers import Number
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
 import pandas as pd
 import sympy as sp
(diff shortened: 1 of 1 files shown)


petab/v2/problem.py
```diff
@@ -1096,6 +1096,46 @@ def __iadd__(self, other):
             )
         return self
 
+    def model_dump(self, **kwargs) -> dict[str, Any]:
+        """Convert this Problem to a dictionary.
+
+        See `pydantic.BaseModel.model_dump <https://docs.pydantic.dev/latest/api/base_model/#pydantic.BaseModel.model_dump>`__
```
dilpath, line 1102: Maybe note that only the tables are round-trip, not the SBML model.

Add a test that it is round-trip? Or is there a different purpose?

dilpath (approved):

petab/v2/problem.py
dweindl: > Or is there a different purpose?

So far, this is mostly to get a quick overview of some (small) PEtab problem without calling each `model_dump` on the attributes individually.
I don't think we need additional full serialization on top of the regular PEtab yaml/tsv format, or do we? `Problem` is pickleable already.
Maybe I should make it private or label it experimental for now?
Will add a comment regarding the model.

matthiaskoenig: The question is how we want to distribute PETab problems via the web. 
Personally I would prefer the COMBINE archive approach, i.e., adding all files to the archive and then sending the archives as files. In this case no full serializability to JSON would be required. The main issue is the serialization of the models.

dweindl: I agree that we should prefer COMBINE archives.

PR #373, opened 2025-04-03
v2: Allow applying multiple conditions simultaneously

dweindl:
Adapt to https://github.com/PEtab-dev/PEtab/pull/616, which allows providing multiple conditions for the same experiment / timepoint.

### DIFF
--- a/petab/v2/core.py
+++ b/petab/v2/core.py
@@ -485,7 +485,7 @@ class ExperimentPeriod(BaseModel):
         alias=C.TIME
     )
     #: The IDs of the conditions to be applied at the start time.
-    condition_ids: list[str] = []
+    condition_ids: list[str] = Field(default_factory=list)
 
     #: :meta private:
     model_config = ConfigDict(populate_by_name=True, extra="allow")
--- a/petab/v2/core.py
+++ b/petab/v2/core.py
@@ -475,7 +475,7 @@ class ConditionTable(BaseModel):
 
 class ExperimentPeriod(BaseModel):
     """A period of a timecourse or experiment defined by a start time
-    and a condition ID.
+    and a list of condition IDs.
 
     This corresponds to a row of the PEtab experiments table.
     """
@@ -484,20 +484,19 @@ class ExperimentPeriod(BaseModel):
     time: Annotated[float, AfterValidator(_is_finite_or_neg_inf)] = Field(
         alias=C.TIME
     )
-    #: The ID of the condition to be applied at the start time.
-    condition_id: str | None = Field(alias=C.CONDITION_ID, default=None)
+    #: The IDs of the conditions to be applied at the start time.
+    condition_ids: list[str] = []
 
     #: :meta private:
     model_config = ConfigDict(populate_by_name=True, extra="allow")
 
-    @field_validator("condition_id", mode="before")
+    @field_validator("condition_ids", mode="before")
     @classmethod
-    def _validate_id(cls, condition_id):
-        if pd.isna(condition_id) or not condition_id:
-            return None
-        if not is_valid_identifier(condition_id):
-            raise ValueError(f"Invalid ID: {condition_id}")
-        return condition_id
+    def _validate_ids(cls, condition_ids):
+        for condition_id in condition_ids:
+            if not is_valid_identifier(condition_id):
+                raise ValueError(f"Invalid ID: {condition_id}")
+        return condition_ids
 
 
 class Experiment(BaseModel):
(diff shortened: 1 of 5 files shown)


petab/v2/core.py
```diff
@@ -548,12 +547,20 @@ def from_df(cls, df: pd.DataFrame) -> ExperimentTable:
...
-                    time=row[C.TIME], condition_id=row[C.CONDITION_ID]
+            periods = []
+            for timepoint in cur_exp_df[C.TIME].unique():
+                condition_ids = [
+                    cid
+                    for cid in cur_exp_df.loc[
+                        cur_exp_df[C.TIME] == timepoint, C.CONDITION_ID
+                    ]
+                    if not pd.isna(cid)
+                ]
```
dilpath, line 558: Might need to add a `dropna`
```suggestion
            for timepoint, cur_exp_time_df in cur_exp_df.groupby(C.TIME):
                condition_ids = cur_exp_time_df[C.CONDITION_ID].unique()
```

petab/v2/lint.py
```diff
@@ -394,6 +397,28 @@ def run(self, problem: Problem) -> ValidationIssue | None:
...
+                if not period.condition_ids:
+                    continue
+                period_targets = set()
+                for condition_id in period.condition_ids:
+                    condition_targets = {
+                        change.target_id
+                        for cond in problem.condition_table.conditions
+                        if cond.id == condition_id
+                        for change in cond.changes
+                    }
```
dilpath, line 413: Is this useful for tools? I guess tools might like some `problem.condition_table.get_updates(condition_id)` that provides a dictionary of updates to be applied. Then the keys can be used here.

petab/v2/lint.py
```diff
@@ -484,11 +509,14 @@ def run(self, problem: Problem) -> ValidationIssue | None:
...
-                if period.condition_id is not None
-            } - available_conditions
+            missing_conditions = (
+                set(
+                    chain.from_iterable(
+                        period.condition_ids for period in experiment.periods
+                    )
+                )
+                - available_conditions
+            )
```
dilpath, line 519: Same thing for here and below, is it useful to add this to the classes? Looks like a lot of additional interface though so fine as is
```python
problem.experiment_table.get_condition_ids()
problem.experiment_table.get_experiment(experiment_id).get_condition_ids()
```

tests/v2/test_lint.py
```diff
@@ -21,3 +21,19 @@ def test_check_experiments():
     tmp_problem = deepcopy(problem)
     tmp_problem["e1"].periods[0].time = tmp_problem["e1"].periods[1].time
     assert check.run(tmp_problem) is not None
+
+
+def test_check_incompatible_targets():
+    """Multiple conditions with overlapping targets cannot be applied
+    at the same time."""
+    problem = Problem()
+    problem.model = SbmlModel.from_antimony("p1 = 1; p2 = 2")
```
dilpath, line 30: :exploding_head:

Didn't know an SBML test model could be specified so simply.

dilpath (approved):
Looks good! Nothing that needs to be changed.

petab/v2/core.py
```diff
@@ -484,20 +484,19 @@ class ExperimentPeriod(BaseModel):
     time: Annotated[float, AfterValidator(_is_finite_or_neg_inf)] = Field(
         alias=C.TIME
     )
-    #: The ID of the condition to be applied at the start time.
-    condition_id: str | None = Field(alias=C.CONDITION_ID, default=None)
+    #: The IDs of the conditions to be applied at the start time.
+    condition_ids: list[str] = []
```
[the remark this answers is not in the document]
dweindl: It's equivalent. Mutable defaults are handled properly.

tests/v2/test_lint.py
dweindl: :muscle:

petab/v2/core.py
dweindl: I wanted to avoid the overhead of the additional groupy / creating the additional dataframes. Usually there will only be a single condition ID.

I'd avoid the `unique` otherwise this will hide errors in case of duplicated conditions.

petab/v2/lint.py
dweindl: If it's already in a dict, the keys will be unique and we can't check for duplicates any. It works if we do the checking every time in such a function.
I think something like that will come when progressing with the importer.

petab/v2/lint.py
dweindl: I am not sure yet. I think, usually one only needs the conditions for a single period, not for the full experiment at once.

petab/v2/lint.py
dilpath: Not that it simplifies the code here, just that it might be useful to tools. But fine to keep as is until a tool dev requests something...

petab/v2/core.py
dilpath: > creating the additional dataframes

Ah, you're right, I thought it was some view instead

> I'd avoid the `unique`

Yes, not sure why I did that...

No issue with your code but I find the groupby faster to understand.
Independently of this, again this might be something that tools want in a library function

Issue #374, opened 2025-04-23
v2: Add new prior distributions


dweindl:
See https://github.com/PEtab-dev/PEtab/pull/595

- [x] Update `petab.v2.core.Parameter` to include the prior distribution
    -> #378 
- [x] Add new `Distribution` classes (requires #335)
     -> #377
- [x] update linter: valid distributions, correct number of parameters, ...
- [x] v1->v2 conversion
    -> #379 
- [x] startpoint sampling via `Problem`
    -> #380

PR #378, opened 2025-04-24
v2: Adapt models to changes in prior distributions

dweindl:
* priorDistribution/priorParameters instead of separate initializationPrior&objectivePrior
* additional prior distributions

Related to https://github.com/PEtab-dev/libpetab-python/issues/374.

### DIFF
--- a/petab/v2/C.py
+++ b/petab/v2/C.py
@@ -191,7 +191,7 @@ OBSERVABLE_TRANSFORMATIONS = [LIN, LOG, LOG10]
 #: Cauchy distribution.
 CAUCHY = "cauchy"
 #: Chi-squared distribution.
-# TODO: "chisquare" in PEtab and sbml-distrib, but usually "chi-squared"
+# FIXME: "chisquare" in PEtab and sbml-distrib, but usually "chi-squared"
 CHI_SQUARED = "chisquare"
 #: Exponential distribution.
 EXPONENTIAL = "exponential"
--- a/petab/v2/C.py
+++ b/petab/v2/C.py
@@ -79,14 +79,10 @@ UPPER_BOUND = "upperBound"
 NOMINAL_VALUE = "nominalValue"
 #: Estimate column in the parameter table
 ESTIMATE = "estimate"
-#: Initialization prior type column in the parameter table
-INITIALIZATION_PRIOR_TYPE = "initializationPriorType"
-#: Initialization prior parameters column in the parameter table
-INITIALIZATION_PRIOR_PARAMETERS = "initializationPriorParameters"
-#: Objective prior type column in the parameter table
-OBJECTIVE_PRIOR_TYPE = "objectivePriorType"
-#: Objective prior parameters column in the parameter table
-OBJECTIVE_PRIOR_PARAMETERS = "objectivePriorParameters"
+#: Prior distribution type column in the parameter table
+PRIOR_DISTRIBUTION = "priorDistribution"
+#: Prior parameters column in the parameter table
+PRIOR_PARAMETERS = "priorParameters"
 
 #: Mandatory columns of parameter table
 PARAMETER_DF_REQUIRED_COLS = [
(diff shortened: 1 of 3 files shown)


petab/v2/C.py
```diff
@@ -193,35 +187,50 @@
...
-LOG_NORMAL = "logNormal"
+#: Log10-normal distribution.
+LOG10_NORMAL = "log10-normal"
 #: Log-Laplace distribution
-LOG_LAPLACE = "logLaplace"
+LOG_LAPLACE = "log-laplace"
+#: Log-normal distribution
+LOG_NORMAL = "log-normal"
+#: Log-uniform distribution.
+LOG_UNIFORM = "log-uniform"
```
dilpath, line 209: Not really in-scope for this PR but is log10-uniform missing? I guess this is how we would convert `parameterScale=log10` into PEtab v2, or is that just no longer possible?

petab/v2/C.py
```diff
@@ -193,35 +187,50 @@
...
-PARAMETER_SCALE_UNIFORM = "parameterScaleUniform"
-#: Normal distribution
-NORMAL = "normal"
-#: Normal distribution on the parameter scale
-PARAMETER_SCALE_NORMAL = "parameterScaleNormal"
+
+#: Cauchy distribution.
+CAUCHY = "cauchy"
+#: Chi-squared distribution.
+# TODO: "chisquare" in PEtab and sbml-distrib, but usually "chi-squared"
```
dilpath, line 194: Seems important to clarify in the spec soon
```suggestion
# FIXME: "chisquare" in PEtab and sbml-distrib, but usually "chi-squared"
```

petab/v2/core.py
```diff
@@ -124,35 +124,41 @@ class NoiseDistribution(str, Enum):
...
-#: Objective prior types as used in the PEtab parameters table.
-ObjectivePriorType = PriorType
-#: Initialization prior types as used in the PEtab parameters table.
-InitializationPriorType = PriorType
 
-assert set(C.PRIOR_TYPES) == {e.value for e in ObjectivePriorType}, (
-    "ObjectivePriorType enum does not match C.PRIOR_TYPES: "
-    f"{set(C.PRIOR_TYPES)} vs { {e.value for e in ObjectivePriorType} }"
+assert set(C.PRIOR_DISTRIBUTIONS) == {e.value for e in PriorDistribution}, (
+    "PriorType enum does not match C.PRIOR_TYPES: "
```
dilpath, line 160:
```suggestion
    "PriorDistribution enum does not match C.PRIOR_DISTRIBUTIONS "
```

petab/v2/problem.py
```diff
@@ -979,22 +973,12 @@ def add_parameter(
...
-                obj_prior_pars = PARAMETER_SEPARATOR.join(
-                    map(str, obj_prior_pars)
-                )
-            record[OBJECTIVE_PRIOR_PARAMETERS] = obj_prior_pars
+        if prior_dist is not None:
+            record[PRIOR_DISTRIBUTION] = prior_dist
+        if prior_pars is not None:
+            if not isinstance(prior_pars, str):
+                prior_pars = PARAMETER_SEPARATOR.join(map(str, prior_pars))
+            record[PRIOR_PARAMETERS] = prior_pars
```
dilpath, line 981:
```suggestion
        if isinstance(prior_pars, Sequence):
            record[PRIOR_PARAMETERS] = PARAMETER_SEPARATOR.join(map(str, prior_pars))
```

dilpath (approved):
:+1:

petab/v2/C.py
dweindl: It's not included in https://github.com/PEtab-dev/PEtab/pull/595. So, so far, it's not available in PEtab v2. Right, I guess we'd need it to upconvert v1 priors. Also log10-laplace. Best to be discussed at https://github.com/PEtab-dev/PEtab.

petab/v2/C.py
dweindl: Agreed.

petab/v2/problem.py
dweindl: I would have relied on the validation in `Parameter`.

dweindl: Oops, prior parameters where missing there altogether...

PR #379, opened 2025-04-25
v1->v2: Update priors

dweindl:
Extend PEtab v1->v2 conversion:

* remove initializationPrior*
* rename objectivePrior*
* convert parameterScale* priors

Related to https://github.com/PEtab-dev/libpetab-python/issues/374

### DIFF
--- a/petab/v2/petab1to2.py
+++ b/petab/v2/petab1to2.py
@@ -416,15 +416,12 @@ def v1v2_parameter_df(
         if pd.isna(pscale):
             pscale = v1.C.LIN
 
-        if (
-            pscale == v1.C.LIN
-            or prior_type not in v1.C.PARAMETER_SCALE_PRIOR_TYPES
-        ):
+        if prior_type not in v1.C.PARAMETER_SCALE_PRIOR_TYPES:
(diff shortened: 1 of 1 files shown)


petab/v2/petab1to2.py
```diff
@@ -393,3 +389,73 @@ def update_noise_dist(row):
...
+
+        pscale = row.get(v1.C.PARAMETER_SCALE)
+        if pd.isna(pscale):
+            pscale = v1.C.LIN
+
+        if (
+            pscale == v1.C.LIN
+            or prior_type not in v1.C.PARAMETER_SCALE_PRIOR_TYPES
+        ):
+            return prior_type
```
dilpath, line 423: For the weird (but valid in v1?) case of `parameterScale=lin` + `objectivePriorType=parameterScale*`, will this return an invalid v2 prior, even though a valid conversion is possible?

dilpath (approved):
:+1:

petab/v2/petab1to2.py
dweindl: You are right, thanks, will fix.

PR #380, opened 2025-04-25
v2: Startpoint sampling

dweindl:
* Add `Parameter.prior_dist`
* Update `v1.distributions.__all__`
* Implement startpoint sampling for `v2.Problem` supporting all new prior distributions

Related to https://github.com/PEtab-dev/libpetab-python/issues/374

### DIFF
--- a/petab/v2/core.py
+++ b/petab/v2/core.py
@@ -25,6 +25,7 @@ from pydantic import (
 )
 from typing_extensions import Self
 
+from ..v1.distributions import *
 from ..v1.lint import is_valid_identifier
 from ..v1.math import petab_math_str, sympify_petab
 from . import C, get_observable_df
@@ -148,6 +149,26 @@ class PriorDistribution(str, Enum):
 assert set(C.PRIOR_DISTRIBUTIONS) == {e.value for e in PriorDistribution}, (
     "PriorDistribution enum does not match C.PRIOR_DISTRIBUTIONS "
     f"{set(C.PRIOR_DISTRIBUTIONS)} vs { {e.value for e in PriorDistribution} }"
+)
+
+_prior_to_cls = {
+    PriorDistribution.CAUCHY: Cauchy,
+    PriorDistribution.CHI_SQUARED: ChiSquare,
+    PriorDistribution.EXPONENTIAL: Exponential,
+    PriorDistribution.GAMMA: Gamma,
+    PriorDistribution.LAPLACE: Laplace,
+    PriorDistribution.LOG10_NORMAL: Normal,
+    PriorDistribution.LOG_LAPLACE: Laplace,
+    PriorDistribution.LOG_NORMAL: Normal,
+    PriorDistribution.LOG_UNIFORM: Uniform,
+    PriorDistribution.NORMAL: Normal,
+    PriorDistribution.RAYLEIGH: Rayleigh,
+    PriorDistribution.UNIFORM: Uniform,
+}
+
+assert not (_mismatch := set(PriorDistribution) ^ set(_prior_to_cls)), (
+    "PriorDistribution enum does not match _prior_to_cls. "
+    f"Mismatches: {_mismatch}"
 )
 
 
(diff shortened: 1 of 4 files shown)


petab/v2/problem.py
```diff
@@ -726,24 +727,29 @@ def get_optimization_to_simulation_parameter_mapping(self, **kwargs):
...
-        See :py:func:`petab.sample_parameter_startpoints`.
+        :returns: The prior distributions for the estimated parameters.
         """
-        return sampling.sample_parameter_startpoints(
-            self.parameter_df, n_starts=n_starts, **kwargs
-        )
+        return {
+            p.id: p.prior_dist
+            for p in self.parameter_table.parameters
+            if p.estimate
```
dilpath, line 738: Seems reasonable to drop non-estimated parameter priors, but then users might see unexpected posterior values. e.g. if they perform MAP then fix the parameters to those values and recompute, there will be an undocumented change in the posterior value because their `estimate` column has changed.
So perhaps priors + fixed parameters should be an error or warning?

petab/v2/core.py
```diff
@@ -929,6 +950,37 @@ def _validate(self) -> Self:
...
+        else:
+            log = False
+
+        if cls == Exponential:
+            if log is not False:
+                raise ValueError(
+                    "Exponential distribution does not support log "
+                    "transformation."
+                )
+            return cls(*self.prior_parameters, trunc=[self.lb, self.ub])
```
dilpath, line 981: Out of curiosity, why explicitly Exponential but not the other distributions for which log- is not supported?

tests/v2/test_problem.py
```diff
@@ -170,3 +171,22 @@ def test_modify_problem():
         }
     ).set_index([PETAB_ENTITY_ID])
     assert_frame_equal(problem.mapping_df, exp_mapping_df, check_dtype=False)
+
+
+def test_sample_startpoint():
```
dilpath, line 176: Unless other asserts are added, like ensuring the samples appear to come from the expected distribution
```suggestion
def test_sample_startpoint_shape():
```

dilpath (approved):
:+1:

petab/v2/core.py
dweindl: In `distributions.py`, I enabled log-versions for all distributions for which I found evidence that they were used anywhere (even if not supported as PEtab prior distribution). Exponential was the only one for which this wasn't the case. Therefore, this is the only one that does not support the `log` argument.

petab/v2/problem.py
dweindl: Hm, I am not sure. It's effectively a different problem then (kind of like changing your prior to a Dirac delta?), so that a change in the posterior should be expected, isn't it?

petab/v2/core.py
dweindl: I think it's closely related to the discussion of whether parameter bounds should truncate the prior distributions, where we concluded that they should. Fixing a parameter is effectively changing the bounds to lb=ub.

petab/v2/problem.py
dilpath: Agreed, a change can be expected, but then priors should be treated as an error since they are dropped, or?

petab/v2/core.py
dilpath: Alright, fair enough! Perhaps add a comment then, since it differs from the spec

petab/v2/problem.py
dilpath: Yes, same issue there, but the worst thing that happens there is that the user looks into their parameter estimates and sees that they can't find the parameter that they thought they estimated. In the priors case, it would be difficult to notice the mistake.

But you're right, and since it's nice to toggle parameters on or off without having to change other columns, fine to not have an error. As you like, but a warning/info/debug for unused priors could still be nice

dweindl: That could be part of some (yet to be implemented) warning mode for validation then, but I'd still find it somewhat confusing to treat priors different than bounds in that situation. Maybe that's something to leave to the parameter estimation tools.

Issue #382, opened 2025-04-28
`petab.v1.simulate.sample_noise` ignores `observableTransformation`


dweindl:
`petab.v1.simulate.sample_noise` silently ignores `observableTransformation`. For example, in case of observableTransformation=log and noiseDistribution=normal, it incorrectly samples from a normal instead of a log-normal distribution.

Reported by @PaulJonasJost

PR #383, opened 2025-04-28
Handle `observableTransformation` in `petab.v1.simulate.sample_noise`

dweindl:
Previously, `petab.v1.simulate.sample_noise` silently ignored `observableTransformation`. For example, in case of observableTransformation=log and noiseDistribution=normal, it incorrectly sampled from a normal instead of a log-normal distribution.

Fixes https://github.com/PEtab-dev/libpetab-python/issues/382.

### DIFF
--- a/petab/v1/simulate.py
+++ b/petab/v1/simulate.py
@@ -256,22 +256,23 @@ def sample_noise(
     observable_transformation = observable_row.get(
         petab.C.OBSERVABLE_TRANSFORMATION, petab.C.LIN
     )
+    transform = lambda x: x
     # observableTransformation=log -> the log of the simulated value is
     #  distributed according to `noise_distribution`
     if observable_transformation == petab.C.LOG:
         simulated_value = np.log(simulated_value)
+        transform = np.exp
     elif observable_transformation == petab.C.LOG10:
         simulated_value = np.log10(simulated_value)
+        transform = lambda x: np.power(10, x) 
 
     # below is e.g.: `np.random.normal(loc=simulation, scale=noise_value)`
     simulated_value_with_noise = getattr(rng, noise_distribution)(
         loc=simulated_value, scale=noise_value * noise_scaling_factor
     )
 
-    if observable_transformation == petab.C.LOG:
-        simulated_value_with_noise = np.exp(simulated_value_with_noise)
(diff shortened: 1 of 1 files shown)


dweindl:
Test failures are unrelated and fixed in https://github.com/PEtab-dev/libpetab-python/pull/381.

petab/v1/simulate.py
```diff
@@ -241,20 +241,38 @@ def sample_noise(
...
+
     # below is e.g.: `np.random.normal(loc=simulation, scale=noise_value)`
     simulated_value_with_noise = getattr(rng, noise_distribution)(
         loc=simulated_value, scale=noise_value * noise_scaling_factor
     )
 
+    if observable_transformation == petab.C.LOG:
+        simulated_value_with_noise = np.exp(simulated_value_with_noise)
+    elif observable_transformation == petab.C.LOG10:
+        simulated_value_with_noise = np.power(10, simulated_value_with_noise)
```
dilpath, line 274: The return value was probably already incorrect, and will be after `np.exp` or `np.power(10, x)` too.

Also moved the untransform/transform logic to the same place, but fine as is if you prefer

```suggestion
    transform = lambda x: x
    # observableTransformation=log -> the log of the simulated value is
    #  distributed according to `noise_distribution`
    if observable_transformation == petab.C.LOG:
        simulated_value = np.log(simulated_value)
        transform = np.exp
    elif observable_transformation == petab.C.LOG10:
        simulated_value = np.log10(simulated_value)
        transform = lambda x: np.power(10, x) 

    # below is e.g.: `np.random.normal(loc=simulation, scale=noise_value)`
    simulated_value_with_noise = getattr(rng, noise_distribution)(
        loc=simulated_value, scale=noise_value * noise_scaling_factor
    )

    # apply observable transformation, ensure `float` type
    simulated_value_with_noise = float(transform(simulated_value_with_noise))
```

dilpath (approved):
Thanks for reporting and fixing, @PaulJonasJost and @dweindl

petab/v1/simulate.py
dweindl: Either works for me.

PaulJonasJost (approved):
Thanks for this 🙏🏼

PR #384, opened 2025-04-28
Allow empty string "" in columns to be overridden with default values in priors

PaulJonasJost:
The check only looked for whether it was not a string, but empty string values that are passed as `""` were not caught here. This should fix the error.

### DIFF
--- a/petab/v1/priors.py
+++ b/petab/v1/priors.py
@@ -278,7 +278,10 @@ class Prior:
         :return: A distribution object.
         """
         dist_type = d.get(f"{type_}PriorType", C.PARAMETER_SCALE_UNIFORM)
-        if not isinstance(dist_type, str) and np.isnan(dist_type):
+        if (
+            (not isinstance(dist_type, str) and np.isnan(dist_type))
+            or not dist_type
+        ):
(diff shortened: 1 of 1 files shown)


petab/v1/priors.py
```diff
@@ -278,7 +278,10 @@ def from_par_dict(
         :return: A distribution object.
         """
         dist_type = d.get(f"{type_}PriorType", C.PARAMETER_SCALE_UNIFORM)
```
dilpath, line 280: IMO the default can come first, and the override can then be stated in the "positive" case, instead of `not`s. This is how it's mostly done elsewhere in the code as I recall.

But fine as is too.

This suggestion is for the changed code too, but github wouldn't let me edit the block.

```suggestion
        dist_type = C.PARAMETER_SCALE_UNIFORM
        if _table_dist_type := d.get(f"{type_}PriorType") and (isinstance(_table_dist_type, str) or not np.isnan(_table_dist_type)):
            dist_type = _table_dist_type
```

dilpath (approved):
Fine as is

dweindl (approved):
Thanks.

PR #385, opened 2025-05-07
Fix parameter scale in sample_from_prior

dweindl:
Return sampled parameters on the parameter scale.
This broke in #329 & #335.


PaulJonasJost (approved):
Works for me!

dilpath (approved):

Issue #390, opened 2025-06-05
v2: Update to new observable placeholder specification


dweindl:
Adapt to the changes in https://github.com/PEtab-dev/PEtab/pull/625

PR #393, opened 2025-07-01
v2: Update to new observable placeholder specification

dweindl:
Adapt to the changes in https://github.com/PEtab-dev/PEtab/pull/625.

Placeholders are now listed explicitly.

Closes https://github.com/PEtab-dev/libpetab-python/issues/390.

### DIFF
--- a/petab/v2/petab1to2.py
+++ b/petab/v2/petab1to2.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import re
 import shutil
 from contextlib import suppress
 from itertools import chain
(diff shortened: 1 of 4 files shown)


petab/v2/petab1to2.py
```diff
@@ -388,6 +391,41 @@ def update_noise_dist(row):
...
+        expr = sympify_petab(formula)
+        # the current sorting will fail for 10+ placeholders, but who does
+        #  that anyway?
+        return v2.C.PARAMETER_SEPARATOR.join(
+            sorted(
+                str(sym)
+                for sym in expr.free_symbols
+                if sym.is_Symbol and pattern.match(str(sym))
+            )
+        )
```
dilpath, line 420: What causes failure for 10+ placeholders?

dilpath (approved):

petab/v2/petab1to2.py
dweindl: Lexicographical sorting:

```
In[1]: list(sorted(["observableParameter1_foo", "observableParameter10_foo"]))
Out[1]: ['observableParameter10_foo', 'observableParameter1_foo']
```

Well, not the sorting itself will fail, but the placeholder order won't match the override order in the measurement table anymore.

Issue #369, opened 2025-03-28
v2: Update validator


dweindl:
So far, validation in petab.v2 is too lax and validator errors aren't too pretty and are often truncated.

* [x] Readable error message
* [x] include sufficient context in error messages
* [ ] Check for missing columns in files
* [ ] Consider using strict mode for validating files, but be more relaxed when programmatically constructing PEtab objects
* [ ] Decide whether we need to be able to de-serialize (somewhat) incorrect PEtab problems

PR #400, opened 2025-07-02
v2: prettify petablint output

dweindl:
Previously, the raw ValidationError tracebacks where shown, which is not what a user wants to see.

Related to #369.


dilpath (approved):
:+1:

PR #401, opened 2025-07-02
Prettify linter output

dweindl:
Prettify linter output in case of schema violations in the problem yaml file. Previously, the messages were rather confusing.

Also fix an error message and a bug in the default schema choice.

Related to #369.


dilpath (approved):
:+1:

PR #410, opened 2025-07-22
v2: Paths as pathlib.Path & validate assignment

dweindl:
For petab.v2 pydantic models, change path attributes to `pathlib.Path`, and validate assignments.


dilpath (approved):
For multiple e.g. measurement tables, are the (relative) paths stored with the table names? If so, is there a method for users to add a table via a path, to a pre-existing list of measurement tables? If not, how should multiple tables be written to disk? Manually specify each individual table filename each write?

dweindl:
We should probably drop `ProblemConfig` from `petab.Problem` and keep only the relevant info there, and then store relative paths in the `*Table` objects. Anything else will be rather inconvenient.

Issue #392, opened 2025-06-05
v2: Handle multi-model problems


dweindl:
PEtab v2 clarifies the semantics for using multiple models inside a single PEtab problem, see https://github.com/PEtab-dev/PEtab/pull/622/files. We need to adapt to that.

This requires:
* [x] #391
* [ ] Multiple models inside `Problem`, or adding a `CompositeProblem` on top of `Problem`
* [ ] Updating many validation rules
* [ ] petab.v2.C.MODEL_ID
* [ ] ...

dweindl:
> * Multiple models inside `Problem`, or adding a `CompositeProblem` on top of `Problem`

I would currently go for changing `petab.v2.Problem.model: Model` to `petab.v2.Problem.models: list[Model]`.

We might want to add some API to split a multi-model problem into separate `petab.v2.Problem`s, where each has only a single model and the associated measurements/conditions/observables/... .

m-philipps:
> We might want to add some API to split a multi-model problem into separate `petab.v2.Problem`s, where each has only a single model and the associated measurements/conditions/observables/... .

Sounds good to me

PR #418, opened 2025-07-24
v2: Basic support for multiple models

dweindl:
Related to #392.

* Let v2.Problem have a list of models
* Support constructing v2.Problem from files with multiple models
* Move some validators to Annotated
* Add some TODOs.

### DIFF
--- a/petab/v2/core.py
+++ b/petab/v2/core.py
@@ -110,6 +110,15 @@ def _valid_petab_id(v: str) -> str:
     return v
 
 
+def _valid_petab_id_or_none(v: str) -> str:
+    """Field validator for optional PEtab IDs."""
+    if not v:
+        return None
+    if not is_valid_identifier(v):
+        raise ValueError(f"Invalid ID: {v}")
+    return v
+
+
 class ParameterScale(str, Enum):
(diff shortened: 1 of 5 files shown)


petab/v2/core.py
```diff
@@ -1481,6 +1472,39 @@ def get_problem(problem: str | Path | Problem) -> Problem:
...
+        if len(self.models) == 0:
+            return None
+
+        raise ValueError(
+            "Problem contains more than one model. "
+            "Use `Problem.models` to access all models."
+        )
+
+    @model.setter
+    def model(self, value: Model):
```
m-philipps, line 1499: Perhaps change the name to make it clear that this modifies the `Problem`, e.g. `set_model`?

m-philipps (approved):
Thank you, nothing important to add from my side.

petab/v2/core.py
dweindl: `model` is a [`property`](https://realpython.com/python-property/). This function here is usually invoked via `problem.model = some_model`, it cannot be invoked as `problem.model(some_model)`. I think the assignment should make it clear that this modifies the object.

PR #419, opened 2025-07-24
Remove parameter mapping from v2.Problem

dweindl:
Leftover from v1. Never worked. Not planned to implement.


m-philipps:
Are you saying it never worked in v1 either?

m-philipps (approved):
Thanks.

PR #421, opened 2025-07-25
Fix PetabStrPrinter for powers

dweindl:
Power expressions with non-atomic bases or exponents were not printed correctly. Fixed here.


m-philipps (approved):
Thanks!

PR #426, opened 2025-08-21
Validate modelId in measurements

dweindl:
Check that we have a valid model ID. Related to https://github.com/PEtab-dev/libpetab-python/issues/392.


m-philipps (approved):
Looks good to me

Issue #424, opened 2025-08-21
Update v2 condition table targetValue validation


dweindl:
After https://github.com/PEtab-dev/PEtab/pull/645, any condition that is used as a first condition in any experiment needs special treatment with regards to allowed symbols in its `targetValue` expression.

PR #427, opened 2025-08-21
Update v2 condition table `targetValue` validation

dweindl:
After https://github.com/PEtab-dev/PEtab/pull/645, any condition that is used as a first condition cannot refer to any symbols other then the parameters listed in the parameter table, or `time'.

Closes https://github.com/PEtab-dev/libpetab-python/issues/424.

### DIFF
--- a/tests/v2/test_lint.py
+++ b/tests/v2/test_lint.py
@@ -82,3 +82,28 @@ def test_undefined_experiment_id_in_measurements():
     problem.measurements[0].experiment_id = "invalid_experiment_id"
     assert (error := check.run(problem)) is not None
     assert "not defined" in error.message
+
+
+def test_validate_initial_change_symbols():
+    """Test validation of symbols in target value expressions for changes
+    applied at the start of an experiment."""
+    problem = Problem()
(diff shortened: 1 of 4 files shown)


tests/v2/test_lint.py
```diff
@@ -39,6 +39,31 @@ def test_check_incompatible_targets():
     assert "overlapping targets {'p1'}" in error.message
 
 
+def test_validate_initial_change_symbols():
+    """Test validation of symbols in target value expressions for changes
+    applied at the start of an experiment."""
+    problem = Problem()
+    problem.model = SbmlModel.from_antimony("p1 = 1; p2 = 2")
+    problem.add_experiment("e1", 0, "c1", 1, "c2")
```
dilpath, line 47: Unrelated: the docstring for `add_experiment` is currently missing that multiple conditions applied at a single timepoint are specified as a list inside `args`

https://github.com/PEtab-dev/libpetab-python/blob/8795dbfe4c5cf42d9f9eae2e88f4b536810cf45d/petab/v2/core.py#L2100-L2109

PR #432, opened 2025-08-25
Fix petablint for v2 problems

### LINKED ISSUE #428: Update for v2 validate_yaml_semantics
PEtab v2 validation works via `problem.validate`, but fails with `petablint` in `validate_yaml_semantics`.
Fix that.

dweindl:
Skip `validate_yaml_semantics` for PEtab v2. Those errors will be caught elsewhere.

Closes #428.

### DIFF
--- a/tests/v2/test_core.py
+++ b/tests/v2/test_core.py
@@ -1,3 +1,4 @@
+import subprocess
 import tempfile
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


tests/v2/test_core.py
```diff
@@ -656,3 +657,43 @@ def test_generate_path():
...
+    problem.add_parameter(
+        "k2", estimate=True, lb=1e-5, ub=1e5, nominal_value=0.5
+    )
+    problem.add_parameter(
+        "sd_A", estimate=True, lb=0.01, ub=10, nominal_value=1
+    )
+    problem.add_measurement(
+        "obs_A", time=10, measurement=2.5, experiment_id=""
+    )
+    assert problem.validate() == []
```
dilpath, line 689: After #430 
```suggestion
    problem.assert_valid()
```

tests/v2/test_core.py
```diff
@@ -656,3 +657,43 @@ def test_generate_path():
...
+
+    problem.config = ProblemConfig(filepath="problem.yaml")
+    problem.models[0].rel_path = "model.xml"
+    problem.parameter_tables[0].rel_path = "parameters.tsv"
+    problem.observable_tables[0].rel_path = "observables.tsv"
+    problem.measurement_tables[0].rel_path = "measurements.tsv"
+    problem.to_files(Path(tmpdir))
+
+    result = subprocess.run(["petablint", str(Path(tmpdir, "problem.yaml"))])  # noqa: S603,S607
+    assert result.returncode == 0
```
dilpath, line 699: Fine as is, but I would normally opt for asserting that more specific output is the same, rather than something that might be coincidentally the same, like a return code of 0. e.g. see the same error in `problem.validate` and `result`.

dilpath (approved):

tests/v2/test_core.py
dilpath: OK, looks good then

PR #452, opened 2025-10-28
v2: validate mappings

dweindl:
https://petab.readthedocs.io/en/latest/v2/documentation_data_format.html#mapping-table:
> The petabEntityId may be the same as the modelEntityId, but it must not be used to alias an entity that already has a valid PEtab identifier. This restriction is to avoid unnecessary complexity in the PEtab problem files.

### DIFF
--- a/petab/v2/core.py
+++ b/petab/v2/core.py
@@ -846,6 +846,22 @@ class Mapping(BaseModel):
         populate_by_name=True, extra="allow", validate_assignment=True
     )
 
+    @model_validator(mode="after")
+    def _validate(self) -> Self:
+        if (
+            self.model_id
+            and self.model_id != self.petab_id
+            and is_valid_identifier(self.model_id)
+        ):
+            raise ValueError(
+                "Aliasing of entities that already have a valid identifier "
+                "is not allowed. Simplify your PEtab problem by removing the "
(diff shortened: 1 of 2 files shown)


dilpath (approved):
> https://petab.readthedocs.io/en/latest/v2/documentation_data_format.html#mapping-table:
> 
> > The petabEntityId may be the same as the modelEntityId, but it must not be used to alias an entity that already has a valid PEtab identifier. This restriction is to avoid unnecessary complexity in the PEtab problem files.

Is there a contradiction here? `petabEntityId` must be a valid PEtab ID. If `petabEntityId` is the same as the `modelEntityId` this implies that the `modelEntityId` is a valid PEtab ID too. But `modelEntityId` must not be a valid PEtab ID. So then `petabEntityId` must differ from `modelEntityId`?

dilpath:
But `x := x` is disallowed because `x` is a valid PEtab ID, right? Hence cannot appear in the `modelEntityId`?

dweindl:
> But `x := x` is disallowed because `x` is a valid PEtab ID, right? Hence cannot appear in the `modelEntityId`?

Not sure.

> The petabEntityId may be the same as the modelEntityId, 

:heavy_check_mark: 

> but it must not be used to alias an entity that already has a valid PEtab identifier. 

My interpretation would be that this doesn't alias anything, so it would be okay.

PR #458, opened 2025-10-31
Add `v2.Problem.__repr__`

dweindl:
Add `v2.Problem.__repr__`, and include ID and all models in `__str__`.

### DIFF
--- a/petab/v2/core.py
+++ b/petab/v2/core.py
@@ -1138,8 +1138,13 @@ class Problem:
         self.mapping_tables = mapping_tables or [MappingTable()]
         self.parameter_tables = parameter_tables or [ParameterTable()]
 
+    def __repr__(self):
+        return f"<{self.__class__.__name__} id={self.id!r}>"
+
     def __str__(self):
-        model = f"with model ({self.model})" if self.model else "without model"
+        pid = repr(self.id) if self.id else "without ID"
+
+        model = f"with models {self.models}" if self.model else "without model"
 
         ne = len(self.experiments)
         experiments = f"{ne} experiments"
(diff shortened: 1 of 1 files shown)


petab/v2/core.py
```diff
@@ -1138,8 +1138,13 @@ def __init__(
         self.mapping_tables = mapping_tables or [MappingTable()]
         self.parameter_tables = parameter_tables or [ParameterTable()]
 
+    def __repr__(self):
+        return f"<{self.__class__.__name__} id={self.id!r}>"
+
     def __str__(self):
-        model = f"with model ({self.model})" if self.model else "without model"
+        pid = repr(self.id) if self.id else "without ID"
```
dilpath: What is the benefit of `repr` here? `self.id` looks like it's either `str` or `None`, so `repr` and `!r` don't change anything?

dilpath (approved):
:+1:

petab/v2/core.py
dweindl: > What is the benefit of `repr` here? `self.id` looks like it's either `str` or `None`, so `repr` and `!r` don't change anything?

As compared to what? Not using `repr` here? The difference will be the quotes.

dilpath: Why not `pid = self.id or "without ID"` since `pid` is used as `f"...{pid}..."` later? Or what do you mean by the difference will be the quotes?

dweindl: > Why not `pid = self.id or "without ID"` since `pid` is used as `f"...{pid}..."` later? Or what do you mean by the difference will be the quotes?

Currently, the model ID is shown in quotation marks if it is not None. With this change, it would be shown unquoted. When additionally using `!r` below, the model ID would be quoted, but so will be the "without ID". (`print("foo", repr("foo"))` -> `foo 'foo'`)

PR #479, opened 2026-05-08
Minor improvement in v1 to v2 converter

m-philipps:
- add Warning when removing `PARAMETER_SCALE`, `INITIALIZATION_PRIOR_TYPE`, `INITIALIZATION_PRIOR_PARAMETERS`
- Use the column order for the experiments table that is suggested in the PEtab format
- Fix description and redundancy in v2 `get_experiment_df`

### DIFF
--- a/petab/v2/petab1to2.py
+++ b/petab/v2/petab1to2.py
@@ -523,7 +523,9 @@ def v1v2_parameter_df(
         errors="ignore",
     )
     # some columns were dropped in PEtab v2
-    if df[v1.C.INITIALIZATION_PRIOR_TYPE].notna():
+    if v1.C.INITIALIZATION_PRIOR_TYPE in df and (
+        df[v1.C.INITIALIZATION_PRIOR_TYPE].notna()
+    ):
         warnings.warn(
             "Initialisation priors in parameter table are not supported "
             "in PEtab v2.",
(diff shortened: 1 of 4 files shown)


petab/v2/petab1to2.py
```diff
@@ -523,6 +523,19 @@ def update_prior(row):
         errors="ignore",
     )
     # some columns were dropped in PEtab v2
+    if v1.C.INITIALIZATION_PRIOR_TYPE in df and (
+        df[v1.C.INITIALIZATION_PRIOR_TYPE].notna().any()
```
dilpath, line 527: Fine for me to drop the `notna` check and warn regardless.

petab/v2/petab1to2.py
```diff
@@ -523,6 +523,19 @@ def update_prior(row):
...
+        warnings.warn(
+            "Initialisation priors in parameter table are not supported "
+            "in PEtab v2.",
+            stacklevel=9,
+        )
+    if not (df[v1.C.PARAMETER_SCALE] == v1.C.LIN).all():
+        warnings.warn(
+            "Parameter scales are not supported in PEtab v2.",
+            stacklevel=9,
+        )
```
dilpath, line 538: Same here, fine for me to warn if the column is present at all

dilpath (approved):
Looks good! Thanks

petab/v2/petab1to2.py
m-philipps: This column is always present in `v1`

PR #480, opened 2026-05-13
Lint mapping table

m-philipps:
- core: Update Fields/Annotations to match the PEtab v2 docs
- add a lint check for the mapping table
- Change the logic in `petab.v2.lint.get_valid_parameters_for_parameter_table`: add mapping table `petabEntityId` regardless of whether the corresponding `modelEntityId` would be valid (was L937->L1036)
- add a test for mapping table linting

### DIFF
--- a/petab/v2/core.py
+++ b/petab/v2/core.py
@@ -19,6 +19,7 @@ from typing import (
     Annotated,
     Any,
     Generic,
+    Literal,
     Self,
     TypeVar,
     get_args,
@@ -318,9 +319,9 @@ class Observable(BaseModel):
     #: Observable name.
     name: str | None = Field(alias=C.OBSERVABLE_NAME, default=None)
     #: Observable formula.
-    formula: sp.Basic | None = Field(alias=C.OBSERVABLE_FORMULA, default=None)
+    formula: sp.Basic = Field(alias=C.OBSERVABLE_FORMULA)
     #: Noise formula.
-    noise_formula: sp.Basic | None = Field(alias=C.NOISE_FORMULA, default=None)
+    noise_formula: sp.Basic = Field(alias=C.NOISE_FORMULA)
     #: Noise distribution.
     noise_distribution: NoiseDistribution = Field(
         alias=C.NOISE_DISTRIBUTION, default=NoiseDistribution.NORMAL
@@ -926,7 +927,7 @@ class Parameter(BaseModel):
     )
     #: Nominal value.
     nominal_value: Annotated[
-        float | None, BeforeValidator(_convert_nan_to_none)
+        float | Literal["array"] | None, BeforeValidator(_convert_nan_to_none)
     ] = Field(alias=C.NOMINAL_VALUE, default=None)
     #: Is the parameter to be estimated?
     estimate: bool = Field(alias=C.ESTIMATE, default=True)
--- a/petab/v2/core.py
+++ b/petab/v2/core.py
@@ -19,7 +19,6 @@ from typing import (
     Annotated,
     Any,
     Generic,
-    Literal,
     Self,
     TypeVar,
     get_args,
@@ -927,7 +926,7 @@ class Parameter(BaseModel):
     )
     #: Nominal value.
     nominal_value: Annotated[
-        float | Literal["array"] | None, BeforeValidator(_convert_nan_to_none)
+        float | None, BeforeValidator(_convert_nan_to_none)
     ] = Field(alias=C.NOMINAL_VALUE, default=None)
     #: Is the parameter to be estimated?
     estimate: bool = Field(alias=C.ESTIMATE, default=True)
--- a/petab/v2/lint.py
+++ b/petab/v2/lint.py
@@ -446,7 +446,7 @@ class CheckUniquePrimaryKeys(ValidationTask):
 
         # check for uniqueness of all primary keys
         counter = Counter(c.id for c in problem.conditions)
-        duplicates = {id_ for id_, count in counter.items() if count > 1}
+        duplicates = sorted(id_ for id_, count in counter.items() if count > 1)
 
         if duplicates:
             return ValidationError(
@@ -454,7 +454,7 @@ class CheckUniquePrimaryKeys(ValidationTask):
             )
 
         counter = Counter(o.id for o in problem.observables)
-        duplicates = {id_ for id_, count in counter.items() if count > 1}
+        duplicates = sorted(id_ for id_, count in counter.items() if count > 1)
 
         if duplicates:
             return ValidationError(
@@ -462,7 +462,7 @@ class CheckUniquePrimaryKeys(ValidationTask):
             )
 
         counter = Counter(e.id for e in problem.experiments)
-        duplicates = {id_ for id_, count in counter.items() if count > 1}
+        duplicates = sorted(id_ for id_, count in counter.items() if count > 1)
 
         if duplicates:
             return ValidationError(
@@ -470,7 +470,7 @@ class CheckUniquePrimaryKeys(ValidationTask):
             )
 
         counter = Counter(p.id for p in problem.parameters)
-        duplicates = {id_ for id_, count in counter.items() if count > 1}
+        duplicates = sorted(id_ for id_, count in counter.items() if count > 1)
 
         if duplicates:
             return ValidationError(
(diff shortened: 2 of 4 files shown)


petab/v2/lint.py
```diff
@@ -977,20 +1079,13 @@ def get_required_parameters_for_parameter_table(
...
+    # Start with mapping table petab ids
+    parameter_ids = {m.petab_id for m in problem.mappings}
 
     # Add parameters from measurement table, unless they are fixed parameters
     def append_overrides(overrides):
         parameter_ids.update(
-            str_p
-            for p in overrides
-            if isinstance(p, sp.Symbol)
-            and (str_p := str(p)) not in condition_targets
```
m-philipps, line 993: Not necessary since this is collected as a set and the `condition_targets` are removed further down.

petab/v2/core.py
```diff
@@ -926,7 +927,7 @@ class Parameter(BaseModel):
     )
     #: Nominal value.
     nominal_value: Annotated[
-        float | None, BeforeValidator(_convert_nan_to_none)
+        float | Literal["array"] | None, BeforeValidator(_convert_nan_to_none)
```
dweindl: Belongs to #482.

petab/v2/lint.py
```diff
@@ -551,7 +554,8 @@ def run(self, problem: Problem) -> ValidationIssue | None:
 
 class CheckAllParametersPresentInParameterTable(ValidationTask):
     """Ensure all required parameters are contained in the parameter table
-    with no additional ones."""
+    with no additional ones. This also ensures that the mapping table petab ids
+    are used in the PEtab problem."""
```
dweindl:
```suggestion
    with no additional ones. 
    
    This also ensures that the mapping table petab ids
    are used in the PEtab problem."""
```

Single sentence for sphinx summaries.

petab/v2/lint.py
```diff
@@ -894,6 +898,78 @@ def run(self, problem: Problem) -> ValidationIssue | None:
...
 
 
+class CheckMappingTable(ValidationTask):
+    """Validate the mapping table."""
+
+    def run(self, problem: Problem) -> ValidationIssue | None:
+        messages = []
+
+        # Mapping table is optional
+        if problem.mappings:
```
dweindl:
```suggestion
        # Mapping table is optional
        if problem.mappings:
            return None
            
        messages = []
        ...
```

Reduce indentation.

petab/v2/lint.py
```diff
@@ -894,6 +898,78 @@ def run(self, problem: Problem) -> ValidationIssue | None:
...
+            # Check that each id, across both the petabEntityId and
+            # modelEntityId columns, occurs only once
+            must_be_unique_ids = []
+            for mapping in problem.mappings:
+                petab_id = getattr(mapping, "petab_id", None)
+                model_id = getattr(mapping, "model_id", None)
```
dweindl: In which situation is this preferable over just  `petab_id = mapping.petab_id`?

petab/v2/lint.py
```diff
@@ -894,6 +898,78 @@ def run(self, problem: Problem) -> ValidationIssue | None:
...
+
+                if petab_id:
+                    must_be_unique_ids.append(petab_id)
+                # Duplicates for annotation-only rows (identity mappings)
+                # are permitted.
```
dweindl: By duplicates, you mean `petab_id == model_id`, not multiple rows with the same `petab_id` or `model_id`, right?
It's not super clear from the specs whether the latter would be allowed or not.

petab/v2/lint.py
```diff
@@ -894,6 +898,78 @@ def run(self, problem: Problem) -> ValidationIssue | None:
...
+            new_petab_ids = {
+                m.petab_id
+                for m in problem.mappings
+                # Ignore identity mappings used for annotation
+                if m.petab_id != m.model_id
+            }
+            old_petab_ids = (
+                {c.id for c in problem.conditions}
+                | {e.id for e in problem.experiments}
+                | {o.id for o in problem.observables}
```
dweindl: Also check for parameter table parameters?

petab/v2/lint.py
```diff
@@ -933,9 +1009,20 @@ def get_valid_parameters_for_parameter_table(
...
     )
 
+    # Add petab ids from mapping table if they are used for aliasing
+    # FIXME only add mapping.petab_id to allowed parameter IDs list if it
+    #       aliases an invalid PEtab ID? See
+    #       https://github.com/PEtab-dev/libpetab-python/pull/482#discussion_r3420762034
     for mapping in problem.mappings:
-        if mapping.model_id and mapping.model_id in parameter_ids.keys():
+        if mapping.petab_id not in invalid:
             parameter_ids[mapping.petab_id] = None
```
dweindl, line 1015: This might add all kinds of entities to the allowed-in-parameter-table dict. Species, observables, experiments, ...
For the currently supported model types (SBML, PySB), I don't think there can be a situation where the mapping table would contribute additional values allowed in the parameter table. For a model type where this could happen, we'd need some API for accessing parameters that are allowed in the parameter table *if they had proper IDs*. This doesn't exist yet. So either we need to add it now, or completely ignore the mapping table in this function. (The same holds for `get_required_parameters_for_parameter_table` below.)

petab/v2/lint.py
```diff
@@ -977,20 +1064,13 @@ def get_required_parameters_for_parameter_table(
...
-    parameter_ids = set()
-    condition_targets = {
-        change.target_id
-        for cond in problem.conditions
-        for change in cond.changes
-    }
+    # Start with mapping table petab ids
+    parameter_ids = {m.petab_id for m in problem.mappings}
 
     # Add parameters from measurement table, unless they are fixed parameters
```
dweindl: With the changes below, this is no longer true.

dweindl (commented):
Thanks. Almost there, I think.

petab/v2/lint.py
dweindl: Wait, that makes no sense. Those aliases can of course be used in the parameter table. Ignore.

petab/v2/lint.py
BSnelling: In case the attribute is not defined, but as both columns are required I think we can assume the attribute will be defined.

petab/v2/lint.py
BSnelling: > a situation where the mapping table would contribute additional values allowed in the parameter table

SciML problems have this use case don't they? For example, we'd define `net_ps` as a PEtab id for the network parameters `net1.parameters` in the mapping table, and then also have `net_ps` specified in the parameters table with its bounds etc.

petab/v2/lint.py
BSnelling: Yes, I've reworded to comment to make this clearer.

petab/v2/lint.py
dweindl: Quite possible, I am not sufficiently familiar with petab-sciml. In that case, I would ignore it for now, revert the changes in `get_valid_parameters_for_parameter_table`, and address it after this PR and #482. It will probably require a bit more thought and changes.

dweindl: The same for `get_required_parameters_for_parameter_table`. As far as I can tell, with the currently (pre-sciml) supported models, the mapping table won't affect what's allowed or required in the parameter table.

BSnelling: I have removed all the mapping table related changes from both functions and updated the unit test.

petab/v2/lint.py
```diff
@@ -1033,7 +1108,12 @@ def append_overrides(overrides):
         if not problem.model.has_entity_with_id(str(p))
     )
 
-    # parameters that are overridden via the condition table are not allowed
+    # Parameters that are overridden via the condition table are not allowed
+    condition_targets = {
+        change.target_id
+        for cond in problem.conditions
+        for change in cond.changes
+    }
```
dweindl:
```suggestion
```

Duplicated, see start of this function.

dweindl (approved):
Thanks, looks good to me, except for that duplication!

petab/v2/lint.py
```diff
@@ -551,7 +554,10 @@ def run(self, problem: Problem) -> ValidationIssue | None:
 
 class CheckAllParametersPresentInParameterTable(ValidationTask):
     """Ensure all required parameters are contained in the parameter table
-    with no additional ones."""
+    with no additional ones.
+
+    This also ensures that the mapping table petab ids
+    are used in the PEtab problem."""
```
dilpath: Where does it do this? There is no `ValidationError` with a message like "Unused mapping table PEtab IDs: ...."

dilpath (approved):
Thanks a lot. Looks good!

petab/v2/lint.py
BSnelling: Good point, it doesn't do this yet as I'm planning to handle mapping/parameter table interaction in the sciml branch. I've removed from the docstring.

PR #482, opened 2026-05-26
Support PEtab SciML problems and enable linting

m-philipps:
WIP


petab/v2/core.py
```diff
@@ -926,7 +938,7 @@ class Parameter(BaseModel):
     )
     #: Nominal value.
     nominal_value: Annotated[
-        float | None, BeforeValidator(_convert_nan_to_none)
+        float | Literal["array"] | None, BeforeValidator(_convert_nan_to_none)
```
dilpath, line 946: Maybe add some `# PEtab SciML supports arrays` comment here, otherwise a non-SciML user/dev might wonder why "array" is permitted

petab/v2/core.py
```diff
@@ -1107,6 +1119,77 @@ def n_estimated(self) -> int:
         return sum(p.estimate for p in self.parameters)
 
 
+class Hybridization(BaseModel):
```
dilpath: Move SciML-specific things to a separate file or subdirectory e.g. `sciml/core.py` and import them here? Just for organization; might help to keep SciML things as separate as possible. e.g. do not `import petab_sciml` here, rather only in the separated file.

Nevermind this if it will lead to circular imports etc.

petab/v2/core.py
```diff
@@ -1133,22 +1216,35 @@ def __init__(
         measurement_tables: list[MeasurementTable] = None,
         parameter_tables: list[ParameterTable] = None,
         mapping_tables: list[MappingTable] = None,
+        neural_networks: list[NNModel] | None = None,
+        hybridization_tables: list[HybridizationTable] | None = None,
+        array_data_files: list[ArrayData] | None = None,
         config: ProblemConfig = None,
     ):
-        from ..v2.lint import default_validation_tasks
+        from ..v2.lint import default_validation_tasks, sciml_validation_tasks
```
dilpath, line 1156: Also this, import `from ..v2.sciml.lint` for example, for separation.

petab/v2/core.py
```diff
@@ -2245,6 +2412,64 @@ def add_experiment(self, id_: str, *args):
             Experiment(id=id_, periods=periods)
         )
 
+    def add_hybridization(self, target_id: str, target_value: str):
+        """Add a SciML hybridization table entry to the problem.
+
+        If there are more than one hybridization tables, the hybridization
```
dilpath:
```suggestion
        If there is more than one hybridization table, the hybridization
```

petab/v2/core.py
```diff
@@ -2245,6 +2412,64 @@ def add_experiment(self, id_: str, *args):
...
+        is added to the last one.
```
dilpath:
```suggestion
        is added to the last table.
```

petab/v2/lint.py
```diff
@@ -894,6 +897,93 @@ def run(self, problem: Problem) -> ValidationIssue | None:
...
+
+            for mapping in problem.mappings:
+                # petabEntityId is not referenced in any model
+                for model in problem.models:
+                    if model.has_entity_with_id(mapping.petab_id):
+                        messages.append(
+                            f"`{mapping.petab_id}` is used in the mapping "
+                            "table and referenced directly in the model "
+                            f"`{model.model_id}`."
+                        )
```
dilpath: The spec currently says that the `petabEntityId` can be the same as a `modelEntityId`, hence the `petabEntityId`s can appear in the model, right? @dweindl what do you think?

> **The petabEntityId may be the same as the modelEntityId**, but it must not be used to alias an entity that already has a valid PEtab identifier. This restriction is to avoid unnecessary complexity in the PEtab problem files.

petab/v2/lint.py
```diff
@@ -894,6 +897,93 @@ def run(self, problem: Problem) -> ValidationIssue | None:
...
+        if culprits := (hyb_target_ids & condition_targets):
+            messages.append(
+                f"Hybridization target ids `{culprits}` are also "
+                "target ids in the condition table."
+            )
+        # NN inputs are not used as target values
+        if culprits := (hyb_target_vals & nn_input_ids):
+            messages.append(
+                "The following neural net inputs were used as target values "
+                f"in the Hybridization tbale: `{culprits}`. Please simplify."
```
dilpath:
```suggestion
                f"in the Hybridization table: `{culprits}`."
```

petab/v2/lint.py
```diff
@@ -933,9 +1023,13 @@ def get_valid_parameters_for_parameter_table(
...
     )
 
+    # Add petab ids from mapping table if they are used for aliasing
     for mapping in problem.mappings:
-        if mapping.model_id and mapping.model_id in parameter_ids.keys():
+        if mapping.petab_id not in invalid:
             parameter_ids[mapping.petab_id] = None
+        # An aliased model id is not a valid parameter id
+        if mapping.model_id and mapping.model_id in parameter_ids:
+            del parameter_ids[mapping.model_id]
```
dilpath: As above, I think it's OK to map a model ID to a PEtab ID in the mapping table as long as it's not an alias, i.e. only in the special case where `petabEntityId == modelEntityId`. I think the reason for this in the spec is for support use of the mapping table for annotation, i.e. via additional user-defined annotation columns.

petab/v2/lint.py
dilpath: There is some discussion of this here https://github.com/PEtab-dev/PEtab/pull/669

dilpath (changes requested):
Thanks! Overall it would be good to try to separate as much of the SciML things into separate files or subdirectory if possible. This will help later on when we design a nicer extension interface for libpetab-python, I think. It will also help with maintenance immediately.

petab/v2/lint.py
BSnelling: I added another clause to the conditional here and below to allow the case where `petabEntityId == modelEntityId`.  Please double check that I've understood the nuance correctly.

petab/v2/core.py
BSnelling: I've split out the sciml specific classes and introduced petab/v2/base.py to resolve circular imports. Let me know what you think. 

There is still a fair amount of sciml related code left in petab/v2/core.py because the sciml fields have been added to the `Problem` class. We could consider an inheritance pattern in the future i.e.
```
class SciMLProblem(Problem):
    ...
```
IMO that would be worth considering if/when we want to support another PEtab extension that requires more fields in `Problem`.

dilpath: Thanks a lot, that was very helpful. And I agree, no need to fully implement now.

petab/v2/lint.py
dilpath: Thanks, looks good

petab/v2/core.py
```diff
@@ -1321,6 +1254,45 @@ def from_yaml(
...
+        )
+
+        array_data_files = (
+            [
+                ArrayDataStandard.load_data(_generate_path(f, base_path))
+                for f in config.extensions[C.SCIML].array_files
+            ]
+            if config.extensions and config.extensions[C.SCIML]
+            else None
+        )
```
dilpath: This is a bit more readable for me

```suggestion
        neural_networks = None
        hybridization_tables = None
        array_data_files = None
        if config.extensions and config.extensions[C.SCIML]:
            # Neural network classes are constructed via pytorch for now to get the
            # proper inputs
            neural_networks = [
                NNModel.from_pytorch_module(
                    NNModelStandard.load_data(
                        _generate_path(
                            file_path=nn_config.location,
                            base_path=base_path,
                        )
                    ).to_pytorch_module(),
                    nn_model_id=nn_id,
                )
                for nn_id, nn_config in (
                    config.extensions[C.SCIML].neural_networks or {}
                ).items()
            ]
    
            hybridization_tables = [
                HybridizationTable.from_tsv(f, base_path)
                for f in config.extensions[C.SCIML].hybridization_files
            ]
    
            array_data_files = [
                ArrayDataStandard.load_data(_generate_path(f, base_path))
                for f in config.extensions[C.SCIML].array_files
            ]
```

petab/v2/core.py
```diff
@@ -1636,6 +1611,28 @@ def id(self, value: str):
...
+        """Combined SciML hybridization tables as DataFrame."""
+        return (
+            HybridizationTable(hybridizations).to_df()
+            if (hybridizations := self.hybridizations)
+            else None
+        )
+
+    @hybridization_df.setter
+    def hybridization_df(self, value: pd.DataFrame):
+        self.hybridization_tables = [HybridizationTable.from_df(value)]
```
dilpath: Add some "Specific to PEtab SciML." comment to all docstrings of methods that are implemented in core PEtab?

dilpath (approved):
Thanks very much! I would be happy to get @dweindl 's approval before merging, since it touches his v2 code a bit.

Currently I don't think `CheckHybridizationTable` is properly tested, but fine for now since it's just a placeholder before we implement all SciML checks.

petab/v2/sciml/__init__.py
dweindl, line 1: I'd prefer moving all extension subpackages to `.ext.{extension_id}` to avoid collisions with other petab modules.

petab/v2/sciml/core.py
```diff
@@ -0,0 +1,127 @@
...
+    AnyUrl,
+    BaseModel,
+    ConfigDict,
+    Field,
+    field_validator,
+)
+
+import petab.v2.C as C
+
+from ...v1.math.sympify import sympify_petab
```
dweindl:
```suggestion
from ..math.sympify import sympify_petab
```

Currently this will import the same, but it's safer in case we separate v1/v2 math processing.

petab/v2/base.py
```diff
@@ -0,0 +1,225 @@
+"""Base classes shared across petab.v2 to avoid circular imports."""
```
dweindl: Hm... `.core` was meant for that already.

petab/v2/C.py
```diff
@@ -258,6 +258,8 @@
 MAPPING_FILES = "mapping_files"
 #: Extensions key in the YAML file
 EXTENSIONS = "extensions"
+#: PEtab SciML extension
+SCIML = "sciml"
```
dweindl:
```suggestion
EXT_ID_SCIML = "sciml"
```
Feels safer. We might have extension IDs that collide with existing constants here.

petab/v2/core.py
```diff
@@ -1321,6 +1254,37 @@ def from_yaml(
             else None
         )
 
+        neural_networks = None
```
dweindl: I am not super happy with adding all that to  `Problem`. I see some of the changes are to invasive to fully isolate them, but for many things it should be possible.
How about having some `Problem.ext.sciml: petab.v2.ext.sciml.SciMlExt`, loading all those things in `SciMlExt.__init__` and accessing all those entities through `problem.ext.sciml.{hybridizations,neural_networks,...}`? Too verbose?
This would also circumvent the circular import issues. Those extension classes would only have to imported at the function level, not module level, and we wouldn't have to move things to base.py.

petab/v2/lint.py
```diff
@@ -894,6 +766,62 @@ def run(self, problem: Problem) -> ValidationIssue | None:
...
+                    for mapping in problem.mappings
+                    for attr in ["petab_id", "model_id"]
+                    if getattr(mapping, attr)
+                ]
+            )
+            non_unique = {id_ for id_, count in counter.items() if count > 1}
+            if non_unique:
+                return ValidationError(
+                    f"Mapping table contains non-unique IDs: {non_unique}."
+                )
```
dweindl: Related to @dilpath's comment below - this would incorrectly flag identity mappings, wouldn't it?

petab/v2/lint.py
```diff
@@ -894,6 +766,62 @@ def run(self, problem: Problem) -> ValidationIssue | None:
...
+
+            # petabEntityId is not defined elsewhere in the PEtab problem
+            petab_ids_mapping = {m.petab_id for m in problem.mappings}
+            defined_petab_ids = (
+                {c.id for c in problem.conditions}
+                | {e.id for e in problem.experiments}
+                | {o.id for o in problem.observables}
+            )
+            if petab_ids_mapping & defined_petab_ids:
+                messages.append(
```
dweindl: I'd say this check should only apply to non-identity-mappings.
The current spec is a bit vague there - is it allowed to use the mapping table for annotations for experiments/observables/...? Afaik this was considered a use case.

petab/v2/lint.py
```diff
@@ -894,6 +766,62 @@ def run(self, problem: Problem) -> ValidationIssue | None:
...
+            # Check that each id only occurs once
+            counter = Counter(
+                [
+                    getattr(mapping, attr)
+                    for mapping in problem.mappings
+                    for attr in ["petab_id", "model_id"]
+                    if getattr(mapping, attr)
+                ]
+            )
+            non_unique = {id_ for id_, count in counter.items() if count > 1}
```
dweindl: Let's use `sorted` or `list` for deterministic error messages.

petab/v2/lint.py
```diff
@@ -894,6 +766,62 @@ def run(self, problem: Problem) -> ValidationIssue | None:
...
+                    "defined in the mapping table but also defined through "
+                    "other PEtab tables."
+                )
+
+            for mapping in problem.mappings:
+                # petabEntityId is not referenced in any model
+                for model in problem.models:
+                    if (
+                        model.has_entity_with_id(mapping.petab_id)
+                        and mapping.petab_id != mapping.model_id
```
dweindl: Swap conditions - the second one is way cheaper to evaluate.

petab/v2/lint.py
dweindl: As above, if we allow using the mapping table for general annotation of petab entities, which I think we do, then the current implementation would incorrectly consider, e.g., observables to be allowed in the parameter table, wouldn't it?

tests/v2/test_lint.py
```diff
@@ -107,3 +107,43 @@ def test_validate_initial_change_symbols():
     problem.parameter_tables[0].parameters.remove(problem["p2"])
     assert (error := check.run(problem)) is not None
     assert "contains additional symbols: {'p2'}" in error.message
+
+
+def test_check_mapping_table():
+    """Test checks related to the mapping table."""
+    problem = Problem()
+    # PySB model from PEtab test suite
```
dweindl:
```suggestion
    # SBML model from PEtab test suite
```
and not really from the test suite?

tests/v2/test_lint.py
```diff
@@ -107,3 +107,43 @@ def test_validate_initial_change_symbols():
...
+    problem.model = SbmlModel.from_antimony("a.mean = 1")
```
dweindl: This doesn't create a parameter with ID `a.mean`, but creates an sbml-distrib construct. Is this intended? If so, please describe the rationale in more detail.
If it's just about creating some model with a petab-incompatible ID, then it would have to be a PySB model. In SBML, we don't have that issue and the mapping table will only ever be used for annotation.

pyproject.toml
```diff
@@ -71,6 +72,9 @@ vis = [
     "seaborn",
     "scipy"
 ]
+sciml = [
+    "petab_sciml @ git+https://github.com/PEtab-dev/petab_sciml.git",
```
dweindl: Both git/github references added here will break PyPI deployment. petab_sciml will have to be installed manually until we can reference a pypi package here.

dweindl (changes requested):

petab/v2/extensions/sciml.py
```diff
@@ -0,0 +1,233 @@
...
+        ArrayDataStandard,
+        NNModel,
+        NNModelStandard,
+    )
+except ModuleNotFoundError:
+    pass
+
+from .. import C
+
+
```
dweindl, line 37: Please set `__all__`

petab/v2/extensions/sciml.py
```diff
@@ -0,0 +1,233 @@
...
+    @classmethod
+    def _sympify(cls, v):
+        if v is None or isinstance(v, sp.Basic):
+            return v
+        if isinstance(v, float) and np.isnan(v):
+            return None
+        return sympify_petab(v)
+
+
+class HybridizationTable:
```
dweindl, line 64: Deriving from `BaseTable` as done for the other tables would reduce some redundancy, but fine as is.

petab/v2/extensions/sciml.py
dweindl, line 1: Add to `doc/modules.rst`?

petab/v2/extensions/sciml.py
```diff
@@ -0,0 +1,233 @@
...
+
+
+class NeuralNetConfig(BaseModel):
+    """A neural net in the PEtab SciML problem configuration."""
+
+    location: AnyUrl | Path
+    pre_initialization: bool
+    format: str
+
+    model_config = ConfigDict(
```
dweindl, line 134:
```suggestion
    #: :meta private:
    model_config = ConfigDict(
```

petab/v2/extensions/sciml.py
```diff
@@ -0,0 +1,233 @@
...
+    #: The PEtab SciML format version.
+    version: str = "0.1.0"
+    #: The paths to the array data files.
+    array_files: list[AnyUrl | Path] = []
+    #: The paths to the hybridization tables.
+    hybridization_files: list[AnyUrl | Path] = []
+    #: The neural network IDs and info.
+    neural_networks: dict[str, NeuralNetConfig] | None = {}
+
+    model_config = ConfigDict(
```
dweindl, line 152:
```suggestion
    #: :meta private:
    model_config = ConfigDict(
```

petab/v2/extensions/sciml.py
```diff
@@ -0,0 +1,233 @@
...
+    """SciML extension runtime state.
+
+    Accessible as ``Problem.extensions.sciml``.
+    """
+
+    def __init__(
+        self,
+        neural_networks: list = None,
+        hybridization_tables: list[HybridizationTable] = None,
+        array_data_files: list = None,
```
dweindl:
```suggestion
        neural_networks: list | None = None,
        hybridization_tables: list[HybridizationTable] = None,
        array_data_files: list  | None = None,
```

petab/v2/core.py
```diff
@@ -1322,6 +1350,45 @@ def from_yaml(
...
+
+            hybridization_tables = [
+                HybridizationTable.from_tsv(f, base_path)
+                for f in config.extensions[C.EXT_ID_SCIML].hybridization_files
+            ]
+
+            array_data_files = [
+                ArrayDataStandard.load_data(_generate_path(f, base_path))
+                for f in config.extensions[C.EXT_ID_SCIML].array_files
+            ]
```
dweindl: Please move to a factory method like `SciMLExt.from_config()`, so this simplifies to something like

```python
sciml = SciMLExt.from_config(config) if config.extensions and config.extensions.get(C.EXT_ID_SCIML) else None
extensions = ProblemExtensions(sciml=sciml)
```

petab/v2/core.py
```diff
@@ -2543,12 +2635,22 @@ def to_yaml(self, filename: str | Path):
...
 
+        for ext_id, d_ext in data[C.EXTENSIONS].items():
+            if ext_id == C.EXT_ID_SCIML:
+                # convert Paths to strings
+                for key in ("array_files", "hybridization_files"):
+                    d_ext[key] = list(map(str, d_ext[key]))
+                for nn in d_ext["neural_networks"]:
+                    d_ext["neural_networks"][nn][C.MODEL_LOCATION] = str(
+                        d_ext["neural_networks"][nn][C.MODEL_LOCATION]
+                    )
```
dweindl: Move to `SciMLConfig.to_yaml`?

petab/v2/lint.py
```diff
@@ -1121,6 +1122,16 @@ def append_overrides(overrides):
...
 
+    hybridization_targets = {
+        hyb.target_id for hyb in problem.extensions.sciml.hybridizations
+    }
+    parameter_ids -= hybridization_targets
+    hybridization_target_values = {
+        str(hyb.target_value)
+        for hyb in problem.extensions.sciml.hybridizations
+    }
+    parameter_ids -= hybridization_target_values
```
dweindl: I am not sure if we should always have a  `SciMLExt` instance at `problem.extensions.sciml`. I'd prefer `None` in case of non-petab-sciml problems. What do you think? In that case, check for `problem.extensions.sciml is not None`.

tests/v2/test_sciml.py
```diff
@@ -0,0 +1,141 @@
...
+from petab.v2.core import *
+from petab.v2.core import ModelFile
+from petab.v2.extensions.sciml import NeuralNetConfig
+from petab.v2.lint import sciml_validation_tasks
+from petab.v2.models.sbml_model import SbmlModel
+
+
+def _get_test_problem():
+    problem = Problem()
+    problem.validation_tasks = sciml_validation_tasks
```
dweindl: Is this still necessary? Doesn't that get set elsewhere?

tests/v2/test_sciml.py
```diff
@@ -0,0 +1,141 @@
...
+        model_files=ConfigDict(
+            {"lv": ModelFile(location="lv.xml", language="sbml")}
+        ),
+        parameter_files=["parameters.tsv"],
+        measurement_files=["measurements.tsv"],
+        observable_files=["observables.tsv"],
+        experiment_files=["experiments.tsv"],
+        mapping_files=["mappings.tsv"],
+        extensions={
+            "sciml": {
```
dweindl: Why not `SciMLConfig(...)`?

dweindl (commented):
Thanks, I think it will be much easier to maintain this way. After slimming down `Problem.from_yaml` this should be good to go.

petab/v2/extensions/sciml.py
BSnelling: Using BaseTable results in a circular import unfortunately.

tests/v2/test_sciml.py
BSnelling: Updated to pass the config to `Problem.__init__`, now the sciml validation gets assigned on construction.

petab/v2/lint.py
```diff
@@ -1014,6 +1015,11 @@ def get_valid_parameters_for_parameter_table(
         if mapping.model_id and mapping.model_id in parameter_ids.keys():
             parameter_ids[mapping.petab_id] = None
 
+    if problem.extensions.sciml is not None:
+        for mapping in problem.mappings:
+            if mapping.petab_id not in invalid:
+                parameter_ids[mapping.petab_id] = None
+
```
BSnelling, line 1022: I added this caveat for now to handle the mapping table contributing to the parameter table in SciML problems. There have been a few discussions of this across recent PRs so I've opened https://github.com/PEtab-dev/libpetab-python/issues/494.

dweindl (approved):
Looks good to me. Thanks.

PR #485, opened 2026-06-09
fix: declare scipy runtime dependency

### LINKED ISSUE #483: Cannot import petab.v2 without scipy installed
Without `scipy` installed on version 0.8.2, the following error occurs when running:

```python
import petab.v2
```

Stacktrace:

```python
Traceback (most recent call last):
  File "<python-input-4>", line 1, in <module>
    import petab.v2
  File ".../petab/v2/__init__.py", line 7, in <module>
    from petab.v1.distributions import *  # noqa: F401, E402
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../petab/v1/distributions.py", line 18, in <module>
    from scipy.stats import (
    ...<8 lines>...
    )
ModuleNotFoundError: No module named 'scipy'
```

I am not a Python expert, but I guess `scipy` should either be listed as a required dependency, or the import should be made lazy so it is only needed when the relevant functionality is used.

xianzuyang9-blip:
Fixes #483.

`petab.v2` imports and re-exports `petab.v1.distributions` at module import time, and `petab.v1.distributions` imports `scipy.stats` at the top level. Because `scipy` was only present in extras, a clean install with the core dependencies could still fail on `import petab.v2` with `ModuleNotFoundError: No module named 'scipy'`.

This moves `scipy` into the core project dependencies so the currently exported distribution API is installed with the package. I removed the duplicate `tests` extra entry; the `vis` extra still keeps its scipy entry because that extra expresses the visualization dependency set.

Verification:
- Reproduced before the change in a venv with all core dependencies except scipy installed: `python -c "import petab.v2"` failed at `petab.v1.distributions` importing `scipy.stats`.
- After the change, `pip install -e .` installs metadata containing core `scipy`, and `python -c "import petab.v2"` succeeds.
- `python -m pytest tests/v1/test_distributions.py`: 19 passed.
- `python -m build --wheel`: succeeded.

I also tried `tests/v2/test_core.py tests/v1/test_distributions.py`; the distributions tests passed, while several v2 tests failed because this local environment could not fetch remote GitHub fixtures and did not install the `antimony` test extra. Those failures are unrelated to the dependency metadata change.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,6 +20,7 @@ dependencies = [
     "jsonschema",
     "antlr4-python3-runtime==4.13.1",
     "pydantic>=2.10",
+    "scipy",
 ]
 license = "MIT"
 authors = [
@@ -38,7 +39,6 @@ tests = [
     "pysb",
     "pytest",
     "pytest-cov",
-    "scipy",
 ]
 quality = [
     "pre-commit",


dweindl:
Obsoleted by #486

PR #486, opened 2026-06-09
Lazy scipy imports

### LINKED ISSUE #483: Cannot import petab.v2 without scipy installed

dweindl:
For now, keep scipy optional, unless probability distributions are really needed.

Fixes #483.

### DIFF
--- a/petab/v1/distributions.py
+++ b/petab/v1/distributions.py
@@ -15,15 +15,10 @@ import abc
 from typing import Any
 
 import numpy as np
-from scipy.stats import (
-    cauchy,
-    chi2,
-    expon,
-    gamma,
-    laplace,
-    norm,
-    rayleigh,
-    uniform,
+
+_SCIPY_IMPORT_ERROR = (
+    "scipy is required for this functionality. "
+    "Install it with: pip install scipy"
 )
 
 __all__ = [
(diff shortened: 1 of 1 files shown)


petab/v1/distributions.py
```diff
@@ -342,6 +332,14 @@ def __init__(
...
         log: bool | float = False,
     ):
+        try:
+            from scipy.stats import norm
+        except ImportError as e:
+            raise ImportError(
+                "scipy is required for this functionality. "
+                "Install it with: pip install scipy"
+            ) from e
+        self._dist = norm
```
dilpath, line 344: Turn into function

dilpath (approved):
👍

petab/v1/distributions.py
dweindl: The current way enables better static analysis which I'd prefer here. I'll dereplicate the message, though.

PR #489, opened 2026-06-20
Fix PetabStrPrinter for non-integer rational exponents

wshlavacek:
`petab_math_str(sympify_petab(x))` is not the identity for square roots and other non-integer rational exponents. For example, `petab_math_str(sqrt(a))` returns `a ^ 1/2`, which `sympify_petab` re-parses as `(a^1)/2 = a/2`. This is a silent mangling of the expression.

The cause is that a non-integer `Rational` exponent, such as `1/2`, is a sympy `Atom` but prints as the multi-token string `1/2`, so the `not exp.is_Atom` guard added in #421 does not parenthesize it. Since `^` binds tighter than `/`, the unparenthesized `a ^ 1/2` then parses as `(a^1)/2`.

This fix parenthesizes a non-integer rational exponent explicitly, so `petab_math_str(sqrt(a)) == "a ^ (1/2)"`, which re-parses correctly. Integer powers and the non-atomic base/exponent cases from #421 are unchanged.

### DIFF
--- a/petab/v1/math/printer.py
+++ b/petab/v1/math/printer.py
@@ -41,9 +41,9 @@ class PetabStrPrinter(StrPrinter):
         str_exp = self._print(exp)
         if not base.is_Atom:
             str_base = f"({str_base})"
-        # A non-integer Rational exponent (e.g. sqrt -> 1/2) is an Atom but prints as the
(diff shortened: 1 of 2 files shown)


dweindl (approved):
Great, thanks!

PR #496, opened 2026-06-29
Fix nan-handling in `Subplot.from_df` checks

dweindl:
Fixes `UserWarning: For plotId plot13 in column xLabel contradictory settings ([nan nan]). Proceeding with first entry (nan).`


dilpath (approved):
Thanks!

PR #498, opened 2026-06-30
Allow for condition ids to be defined inside array files (SciML problems only)

BSnelling:
Condition specific inputs to neural networks can be provided as array files for SciML problems. For these condition specific inputs, condition ids referenced in the experiments table need to be defined in the array files rather than the conditions table. https://petab-sciml.readthedocs.io/latest/format.html#inputs

This PR adds:

- a check for condition ids in array files to the `CheckExperimentConditionsExist` validation task
- a workaround for the array-defined condition ids in `_convert_experiment` (AMICI's sciml support handles condition specific array inputs elsewhere)

### DIFF
--- a/petab/v2/converters.py
+++ b/petab/v2/converters.py
@@ -249,6 +249,12 @@ class ExperimentsToSbmlConverter:
             #  single-period experiments.
             if i_period == 0:
                 exp_ind_id = self.get_experiment_indicator(experiment.id)
+                # If condition ids are set by array data, skip as importers handle this
+                if set(period.condition_ids).issubset(
+                    self._new_problem._get_array_data_condition_ids()
+                ):
+                    continue
+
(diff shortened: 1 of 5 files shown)


petab/v2/converters.py
```diff
@@ -249,6 +249,14 @@ def _convert_experiment(self, experiment: Experiment) -> None:
             #  single-period experiments.
             if i_period == 0:
                 exp_ind_id = self.get_experiment_indicator(experiment.id)
+                # Skip if condition ids are set by array data
+                # importers handle this
+                if self._new_problem.extensions.sciml:
+                    if set(period.condition_ids).issubset(
+                        self._new_problem.extensions.sciml._get_array_data_condition_ids()
```
dweindl: Call once and store?

dweindl (approved):

PR #506, opened 2026-08-10
Fix PEtab v2 extension config model

dweindl:
## Summary
- `ProblemConfig.extensions` was typed as `list[ExtensionConfig] | dict`, but only the dict branch was ever used/supported downstream (#474).
- The extension config models didn't match the v2 schema: neither `ExtensionConfig` nor `SciMLConfig` had the schema-mandated `required` field, and the generic `ExtensionConfig` wrongly nested extra keys under a `config` field instead of allowing them directly alongside `version`/`required`. As a result, writing a problem with a `sciml` extension to YAML and reading it back failed schema validation.
- Added a shared `ExtensionConfig` base (`version`, `required`, arbitrary extra fields allowed) in `petab/v2/extensions/__init__.py`; `SciMLConfig` now subclasses it (`required` defaults to `True`, since a SciML hybrid model is virtually always load-bearing for the problem's interpretation).
- `ProblemConfig.extensions` is now `dict[str, ExtensionConfig]` (using `SerializeAsAny` so subclass-specific fields survive serialization).
- Fixed a broken import in `SciMLConfig.to_yaml()` that made it crash unconditionally.
- `Problem.validate()` still only warns about unsupported extensions — rejecting a problem based on an extension's `required` flag is left to consumers (e.g. simulators) that actually interpret the extension mathematically; libpetab-python itself doesn't.

Closes #474

## Test plan
- [x] `python -m pytest tests/v2/` — 106 passed, 1 skipped
- [x] `pre-commit run --all-files` on changed files — ruff/ruff-format clean
- [x] Added regression test round-tripping a `sciml`-extension problem through `ProblemConfig.to_yaml()` + schema validation
- [x] Added tests for generic (non-sciml) extension parsing, rejection of non-dict `extensions`, and unsupported-extension warning severity

🤖 Generated with [Claude Code](https://claude.com/claude-code)


petab/v2/core.py
```diff
@@ -1982,6 +1984,11 @@ def validate(
             and self.config.extensions
             and (self.config.extensions.keys() - supported_extensions)
         ):
+            # Note: whether rejecting a problem that uses an unsupported
+            # extension marked `required` is up to the consumer (e.g. a
+            # simulator) that actually interprets the extension
+            # mathematically -- libpetab-python itself doesn't, so it only
+            # warns that it can't fully lint the problem.
```
BSnelling: I don't think this comment is needed, the warning message below explains the circumstance here.

BSnelling (approved):

petab/v2/extensions/sciml.py
```diff
@@ -136,11 +137,15 @@ class NeuralNetConfig(BaseModel):
...
-class SciMLConfig(BaseModel):
+class SciMLConfig(ExtensionConfig):
     """The extended configuration of a PEtab SciML problem."""
 
     #: The PEtab SciML format version.
     version: str = "0.1.0"
+    #: Whether the extension is required for the mathematical
+    #: interpretation of the problem. Defaults to ``True`` since a SciML
+    #: problem's hybrid ODE/ML model is virtually always load-bearing.
+    required: bool = True
```
dilpath: Since this is just changing inherited values.
```suggestion
    version = "0.1.0"
    required = True
```

petab/v2/core.py
```diff
@@ -2553,23 +2548,26 @@ class ProblemConfig(BaseModel):
...
-                        if isinstance(ext_config, ExtensionConfig)
-                        else ExtensionConfig(**ext_config)
-                    )
-            return parsed_extensions
-        return v
+        if not isinstance(v, dict):
+            raise ValueError(
+                "extensions must be a dict of extension ID to extension "
+                f"config, got {type(v)}."
+            )
```
dilpath, line 2552:
```suggestion
        if not isinstance(v, dict):
            raise TypeError(
                "extensions must be a dict of extension ID to extension "
                f"config, got {type(v)}."
            )
```

petab/v2/core.py
```diff
@@ -2553,23 +2548,26 @@ class ProblemConfig(BaseModel):
...
+                    ext_config
+                    if isinstance(ext_config, SciMLConfig)
+                    else SciMLConfig(**ext_config)
+                )
+            else:
+                parsed_extensions[ext_name] = (
+                    ext_config
+                    if isinstance(ext_config, ExtensionConfig)
+                    else ExtensionConfig(**ext_config)
+                )
```
dilpath: Hm, could all of this be moved to a function inside `petab.v2.extensions` so that there is no explicit dependency on specific extensions here?

tests/v2/test_sciml.py
```diff
@@ -157,6 +157,26 @@ def test_lint():
     assert problem.validate() == []
 
 
+def test_sciml_config_yaml_round_trip(tmp_path):
+    """The `sciml` extension config, once written to YAML via
+    `ProblemConfig.to_yaml()`, is schema-valid and can be read back.
+    """
+    from petab.v1.yaml import load_yaml, validate_yaml_syntax
```
dilpath, line 164: Strange to see `petab.v1` here but fine if intended.

dilpath (approved):
Thanks!

tests/v2/test_sciml.py
dweindl: I am fine with moving them to a common module, but not in this PR.

petab/v2/core.py
dweindl: Pydantic converts `ValueError` to `ValidationError`, but not `TypeError`.

petab/v2/core.py
dweindl: Moved.

PR #507, opened 2026-08-10
Replace tox and pip with uv

dweindl:
## Why
`tox` and `pip` split dependency/environment management across two tools:
`pip` installs packages, `tox` orchestrates per-purpose virtualenvs
(quality/unit/doc) on top of it. `uv` replaces both with a single fast
tool that handles dependency resolution/installation, Python version
management, virtualenv creation, and building sdists — fewer moving
parts for contributors to learn, and noticeably faster CI installs.

## Summary
- Replace `tox` + `pip` with `uv` across dev tooling, CI, and the release workflow.
- CI (`ci_tests.yml`): `astral-sh/setup-uv` installs uv + the matrix Python version (with built-in caching); `tox -e quality` -> `uvx pre-commit run --all-files`; `tox -e unit` -> `uv sync --group ci` + `uv run pytest`. The `"3.x"` matrix wildcard is replaced with an explicit `"3.14"`, since `setup-uv` needs exact versions.
- Release (`deploy.yml`): `pip install setuptools wheel build && python -m build -s` -> `uv build --sdist`.
- The two CI-only git dependencies previously listed in `tox.ini`'s `unit` env `deps=` (`petabtests`, `benchmark_models_petab`), plus the `tests`/`reports`/`combine`/`vis` extras, now live in a single PEP 735 `[dependency-groups]` `ci` group in `pyproject.toml` (via a self-referential `petab[tests,reports,combine,vis]` entry) — confirmed dependency-groups are excluded from built distribution metadata, so PyPI's rejection of direct/VCS references doesn't apply here.
- No `uv.lock` is committed — this is a library, so CI keeps resolving against current constraints each run, matching the previous un-pinned tox/pip behavior.
- Removed `tox.ini` and `requirements-dev.txt` (only contained `tox>=3.26.0`).
- Updated `doc/development.rst` dev-workflow commands to use `uv sync`/`uv run`/`uvx`.
- Fixed `README.md`'s stated Python requirement (`>=3.11` -> `>=3.12`) to match `pyproject.toml`'s `requires-python`.
- Added `[tool.uv] constraint-dependencies = ["pysb>=1.17.0"]`: uv resolves one dependency graph across all declared extras (not just the ones requested for a given sync), so the unit-test job was transitively pulling in `pysb` via the unrelated `doc` extra and landing on `1.16.0`, whose legacy build bootstrap hits a dead URL. Pinning the resolver to `>=1.17.0` avoids it without changing the public `tests`/`doc` extras' declared ranges.

## Not changed (flagged, out of scope)
- `.readthedocs.yaml` stays on `pip` — Read the Docs' `python.install` only supports `pip`/`setuptools` natively; using `uv` there needs custom `build.commands` overrides.
- Contributors lose the single `tox -e <name>` entrypoint in favor of direct `uv run`/`uvx` commands.

## Test plan
- [x] `uv build --sdist` produces an equivalent sdist to the previous `python -m build -s`.
- [x] `uv sync --group ci` resolves and installs correctly (verified `petabtests` and `benchmark_models_petab` install from git, and the full test suite passes locally: 378 passed, 4 skipped).
- [x] `uvx pre-commit run --all-files` behaves like the old `tox -e quality`.
- [x] CI (`ci_tests.yml`) passes across the full OS x Python-version matrix on this PR.
- [ ] `deploy.yml` succeeds on the next release (can't be tested pre-merge).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


.github/workflows/ci_tests.yml
```diff
@@ -12,47 +12,32 @@ jobs:
...
     - name: Quality tests
-      run: tox -e quality
+      run: uvx pre-commit run --all-files
       if: matrix.platform == 'ubuntu-latest'
 
     - name: Unit tests
-      run: tox -e unit
+      run: |
+        uv sync --group ci
+        uv pip install "sympy>=1.12.1"
```
dilpath: Move this dependency to e.g. pyproject.toml?

dilpath (approved):
👍

.github/workflows/ci_tests.yml
dweindl: Not yet possible. Still no pysb release on pypi that would be compatible with that sympy requirement.

dweindl: Implemented via `override-dependencies`

PR #501, opened 2026-07-05
Add support for BNGL models

wshlavacek:
Adds BNGL as a PEtab model language next to SBML and PySB, so a problem
declaring `language: bngl` loads and validates instead of stopping at
model-load in `petablint`.

`BnglModel` reads BNGL with a pure-Python parser that pulls the entities
validation cares about (parameters, species, observables, functions) straight
from the model text. So loading and validating a BNGL problem needs no external
tools and adds no runtime dependency. `is_valid()` will additionally run
`BNG2.pl --check` if that executable is on the `PATH`, for a stricter check, and
returns `True` if it isn't — so BNG2.pl is used when available but never
required. Both v1 and v2 are covered.

PyBNF has been dogfooding this reader for a while — it uses it to fit rule-based
models against PEtab problems, and its PEtab tutorials exercise the loader end
to end: https://github.com/lanl/PyBNF/tree/main/examples/tutorial

Addresses PEtab-dev/PEtab#436.

PR #508, opened 2026-08-12
Add support for BNGL models (cont.)

dweindl:
Builds on #501 (BNGL model support, by @wshlavacek - full credit for the reader implementation and model corpus selection, all of which this branch carries forward unchanged).

* Removes most of the vendored `.bngl` files and downloads them on demand


tests/v1/bngl_corpus/README.md
```diff
@@ -0,0 +1,36 @@
...
+Public community BNGL models, chosen for feature coverage, backing
+`tests/v1/test_bngl_corpus.py`: it asserts `parse_bngl` enumerates the same
+model entities BNG2.pl does. The BNG2.pl answers are cached in `golden.json`
+(the entity name sets BNG2.pl emits from `writeModel`, its canonical parse),
+so the test needs **no BNG2.pl** -- it compares the reader against the frozen
+oracle. Regenerate the golden (needs BNG2.pl) with:
+
+    python tests/v1/test_bngl_corpus.py
+
+Vendored from public repos (RuleWorld/RuleHub, wshlavacek/BNGL-Models).
```
dilpath: Links

tests/v1/fetch_bngl_corpus.py
```diff
@@ -0,0 +1,324 @@
...
+        "Published/An2009/An_2009.bngl",
+        "1b48efd6fc190988faa5eb28fed78af492f49f8fdf67f064436fbf0f14b634e9",
+        None,
+    ),
+    (
+        "Barua_2009__PATCHED.bngl",
+        _RULEHUB,
+        "Published/Barua2009/Barua_2009.bngl",
+        "26ca5053c4a340b597b2d839edd736469fc14cc2a03cf96b0447b7537089c454",
+        ("atoll=>", "atol=>", 1),
```
dilpath, line 64: Should rather be an rulehub PR? Then remove the "repair" feature.

tests/v1/fetch_bngl_corpus.py
```diff
@@ -0,0 +1,324 @@
...
+import sys
+from pathlib import Path
+
+# Running this file directly makes the interpreter prepend its own directory
+# to sys.path. The `math/` directory then shadows the stdlib `math` module,
+# resulting in an ImportError. Drop that entry before importing anything that
+# could pull in `math`.
+_SCRIPT_DIR = str(Path(__file__).resolve().parent)
+if sys.path and sys.path[0] == _SCRIPT_DIR:
+    del sys.path[0]
```
dilpath, line 31: Clean up PATH after its modification is no longer needed, instead of here?

dilpath (approved):
Overall, fine for me since it doesn't affect much pre-existing code.

Will there be example PEtab problems with BNGL models here in the tests, or in the petab_test_suite? 

Or some guide for BNGL users to describe how e.g. only species exported via BNGL observables are valid for PEtab observable formulae?

wshlavacek:
There are PEtab lessons included in the PyBNF tutorial:
https://github.com/lanl/PyBNF/tree/main/examples/tutorial/

tests/v1/fetch_bngl_corpus.py
dweindl: I am not familiar with the policy there, but I assume it's deliberately untouched there.

tests/v1/fetch_bngl_corpus.py
dweindl: I am not sure what you mean. This is not about the `PATH` environment variable. It's Python's import machinery and our directory names here.

dweindl:
> Will there be example PEtab problems with BNGL models here in the tests, or in the petab_test_suite?

Ideally yes, but I don't see that as a requirement for adding BNGL support here.

> Or some guide for BNGL users to describe how e.g. only species exported via BNGL observables are valid for PEtab observable formulae?

Some additional model-type specific information would be valuable, yes. However, I don't think that belongs here - I would add that to https://github.com/PEtab-dev/PEtab.

PR #509, opened 2026-08-12
Fix SBML math parsing

### LINKED ISSUE #283: Fix SBML math parsing
`petab.models.sbml_model.sympify_sbml` currently uses `sympy.sympify` for evaluating SBML L3 formulas. Operator precedence is different between sympy and SBML L3 formulas. This can lead to hard-to-find issues. Use a proper parser, e.g. https://github.com/dweindl/sbmlmath/.

dweindl:
- `sympify_sbml` now parses SBML MathML directly via `sbmlmath.sbml_math_to_sympy` instead of round-tripping through `libsbml.formulaToL3String` + `sympy.sympify`, which silently applies Python operator semantics to SBML L3 formulas.
- Promoted `sbmlmath` from the `doc` extra to a core dependency, since `petab.v1` now needs it unconditionally.
- Added a regression test: SBML's `%` uses sign-of-dividend semantics, Python's `%` uses sign-of-divisor — the old code silently returned the wrong value for negative operands.

Fixes #283


tests/v1/test_sbml.py
```diff
@@ -150,3 +150,22 @@ def test_sbml_from_to_ant():
...
+    sbml_document = libsbml.SBMLDocument(3, 2)
+    sbml_model = sbml_document.createModel()
+    parameter = sbml_model.createParameter()
+    parameter.setId("e")
+    parameter.setConstant(False)
+    initial_assignment = sbml_model.createInitialAssignment()
+    initial_assignment.setSymbol("e")
+    initial_assignment.setMath(libsbml.parseL3Formula("-5 % 3"))
+
+    assert sympify_sbml(initial_assignment).doit() == -2
```
dilpath, line 171: Move to `sbmlmath`?

dilpath (approved):
👍

tests/v1/test_sbml.py
dweindl: It can be added there as well, but I think it makes sense to have it here. This tests exactly what changed in this PR. Would have failed before, passes now.

PR #510, opened 2026-08-12
Precommit upgrade

dweindl:
Addressing unparenthesized implicit string concatenation, implicit `Optional`, `subprocess.run` without explicit `check` argument and some others.


dilpath (approved):
👍

PR #511, opened 2026-08-13
SciML linting - allow nn outputs not in params table

BSnelling:
Allow network outputs to be used in observable and noise formulae without having to appear in the parameters table. The PR updates the `get_required_parameters_for_parameter_table` check to reflect this.

### DIFF
--- a/petab/v2/lint.py
+++ b/petab/v2/lint.py
@@ -1127,6 +1127,8 @@ def get_required_parameters_for_parameter_table(
     parameter_ids -= condition_targets
 
     if problem.extensions.sciml is not None:
+        from .extensions.sciml_lint import get_nn_entity_petab_ids
+
         hybridization_targets = {
             hyb.target_id for hyb in problem.extensions.sciml.hybridizations
         }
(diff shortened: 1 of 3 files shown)


petab/v2/lint.py
```diff
@@ -1137,6 +1137,16 @@ def append_overrides(overrides):
         }
         parameter_ids -= hybridization_target_values
 
+        # NN outputs can be used in observable and noise formulas without
+        # appearing in the parameter table.
+        try:
+            from .extensions.sciml_lint import get_nn_entity_petab_ids
+        except ImportError:
+            pass
```
dweindl: This `pass` doesn't feel safe. I think `get_nn_entity_petab_ids` should only be imported/called if `problem` required SciML, but then, any import failure should be an error.

dweindl (commented):
Thanks, good after gating on SciML-usage and not hiding the import error.

petab/v2/lint.py
BSnelling: Makes sense, I've updated accordingly.

dweindl (approved):
👍

dilpath (approved):
Looks good but I think the wording is a bit suboptimal. NN outputs should never appear in the parameter table, right? But currently the PR seems to imply that there are cases where it is OK

BSnelling:
> Looks good but I think the wording is a bit suboptimal. NN outputs should never appear in the parameter table, right?

I thought this was the case, but wasn't 100% sure. I have updated the wording now to say they should not appear.

tests/v2/test_sciml.py
```diff
@@ -401,6 +401,57 @@ def test_parameter_posterior_requires_bounds_or_prior():
...
+    problem.add_observable("fitness_obs", "net1_output2", noise_formula="0.05")
+    problem.add_measurement(
+        "fitness_obs", time=1, measurement=1, experiment_id="e1"
+    )
+    return problem
+
+
+def test_nn_output_in_observable_formula_not_required_parameter():
+    """NN outputs consumed by an observable formula are not parameter table
+    entries -- PEtab SciML explicitly allows an NN output in a formula."""
```
dilpath:
```suggestion
    """NN outputs should not appear in the parameter table, and can appear in observable formulas."""
```

